PR #58, opened 2015-07-16
Update the pip recommendations

dstufft:
* Recommend at least pip 7 because versions since 1.3 have changed the defaults to better protect end users during install.
* Use the new official location for the bootstrap script.


hellais:
Looks good to me.

Thanks!

PR #62, opened 2016-02-01
Feature/json

hellais:
This is not ready to be merged, but is a PR associated with: https://github.com/TheTorProject/ooni-probe/pull/457 https://github.com/TheTorProject/ooni-probe/pull/456 and https://github.com/TheTorProject/ooni-spec/pull/49

### DIFF
--- a/oonib/otime.py
+++ b/oonib/otime.py
@@ -95,5 +95,5 @@ def timestamp(t=None):
     """
     if t is None:
         t = datetime.utcnow()
-    ISO8601 = "%Y-%m-%dT%H%M%SZ"
+    ISO8601 = "%Y%m%dT%H%M%SZ"
     return t.strftime(ISO8601)
--- a/oonib/report/handlers.py
+++ b/oonib/report/handlers.py
@@ -18,6 +18,11 @@ from oonib.config import config
 
 def report_file_name(archive_dir, report_details):
     timestamp = datetime.fromtimestamp(report_details['start_time'])
+    ext = report_details.get("format")
+    if ext == "json":
+        ext = "json"
+    else:
+        ext = "yamloo"
     keys = dict(
         report_details.items(),
         iso8601_timestamp=otime.timestamp(timestamp),
@@ -26,9 +31,10 @@ def report_file_name(archive_dir, report_details):
         day=timestamp.strftime("%d"),
         hour=timestamp.strftime("%H"),
         minute=timestamp.strftime("%M"),
-        second=timestamp.strftime("%S")
+        second=timestamp.strftime("%S"),
+        ext=ext
     )
-    report_file_template = "{probe_cc}/{test_name}-{iso8601_timestamp}-{probe_asn}-probe.yamloo"
+    report_file_template = "{probe_cc}/{test_name}-{iso8601_timestamp}-{probe_asn}-probe.{ext}"
     if config.main.report_file_template:
         report_file_template = config.main.report_file_template
     dst_filename = os.path.join(archive_dir, report_file_template.format(**keys))
@@ -45,13 +51,14 @@ class Report(object):
                  stale_time,
                  report_dir,
                  archive_dir,
-                 reports):
+                 reports, file_format="yaml"):
         self.report_id = report_id
 
         self.stale_time = stale_time
         self.report_dir = report_dir
         self.archive_dir = archive_dir
         self.reports = reports
+        self.file_format = file_format
 
         self.refresh()
 
@@ -76,13 +83,23 @@ class Report(object):
         report_filename = get_report_path(self.report_id)
         try:
             with open(report_filename) as fd:
-                g = yaml.safe_load_all(fd)
-                report_details = g.next()
+                if self.file_format == "json":
+                    line = fd.readline()
+                    json.loads(line)
+                else:
+                    g = yaml.safe_load_all(fd)
+                    report_details = g.next()
         except IOError:
             raise e.ReportNotFound
 
         dst_filename = report_file_name(self.archive_dir, report_details)
         shutil.move(report_filename, dst_filename)
+
+        if self.file_format == "json":
+            report_details["record_type"] = "footer"
+            with open(dst_filename, "a+") as fd:
+                json.dump(report_details, dst_filename)
+                fd.write("\n")
 
         if not self.delayed_call.called:
             self.delayed_call.cancel()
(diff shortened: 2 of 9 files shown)


oonib/otime.py
```diff
@@ -95,5 +95,5 @@ def timestamp(t=None):
     """
     if t is None:
         t = datetime.utcnow()
-    ISO8601 = "%Y-%m-%dT%H%M%SZ"
+    ISO8601 = "%Y%m%dT%H%M%SZ"
```
bassosimone: I like this. Wondering, however, if this could cause backwards compatibility issues?

oonib/report/handlers.py
```diff
@@ -75,13 +88,13 @@ def get_report_path(report_id):
...
-                g = yaml.safe_load_all(fd)
-                report_details = g.next()
+            open(report_filename).close()
         except IOError:
             raise e.ReportNotFound
 
-        dst_filename = report_file_name(self.archive_dir, report_details)
+        dst_filename = report_file_name(archive_dir=self.archive_dir,
+                                        report_details=self.report_details,
+                                        report_id=self.report_id)
```
bassosimone: I find it confusing that there is a variable named `report_filename` and a function named `report_file_name`

oonib/report/handlers.py
```diff
@@ -126,63 +138,58 @@ def parseNewReportRequest(request):
...
+        header['probe_asn'] = 'AS0'
+
+    if not header.get('probe_cc'):
+        header['probe_cc'] = 'ZZ'
+
+    if not header.get('start_time'):
+        header['start_time'] = time.time()
+
+    if not header.get('start_time'):
+        header['start_time'] = time.time()
```
bassosimone: These two lines duplicate two equal lines above them

oonib/report/handlers.py
```diff
@@ -126,63 +138,58 @@ def parseNewReportRequest(request):
...
+
+    if not header.get('data_format_version'):
+        header['data_format_version'] = '0.1.0'
```
bassosimone: Where 0.1.0 indicates yaml, right?

oonib/test/test_report.py
```diff
@@ -206,5 +187,3 @@ def test_create_update_and_close_report(self):
             for i in range(report_entry_count):
                 self.assertEqual(yaml.safe_load(sample_report_entry),
                                  written_report.next())
-
-        self.directories.add('ZZ')
```
bassosimone: Wouldn't it more accurate to add a test case in which you update a report by pushing a json entry?

oonib/testhelpers/http_helpers.py
```diff
@@ -168,6 +172,14 @@ def all(self, length, keyword):
...
         self.write(self.genRandomPage(length, keyword))
 
+class TCPConnectProtocol(Protocol):
+    def connectionMade(self):
+        self.transport.loseConnection()
+
+class TCPConnectFactory(Factory):
+    def buildProtocol(self, addr):
+        return TCPConnectProtocol()
+
```
bassosimone: Are these two new classes used in this diff?

oonib/report/handlers.py
```diff
@@ -101,24 +114,23 @@ def parseUpdateReportRequest(request):
...
-    except KeyError:
+
+    report_id = parsed_request.get('report_id', report_id)
+    if not report_id:
         raise e.MissingField('report_id')
 
     if not re.match(report_id_regexp, report_id):
         raise e.InvalidRequestField('report_id')
 
+    parsed_request['report_id'] = report_id
```
bassosimone: Here, aren't you overwriting `report_id` with `report_id`? That is, you have excluded the case in which the report id read from `parsed_request` was not set and the one in which was invalid, so you probably know it is valid and present at this point, so why you overwrite it?

oonib/report/handlers.py
```diff
@@ -126,63 +138,58 @@ def parseNewReportRequest(request):
...
+            value_to_check = header[k]
         except KeyError:
             raise e.MissingField(k)
 
-        print "Matching %s with %s | %s" % (regexp, value_to_check, k)
+        log.debug("Matching %s with %s | %s" % (regexp, value_to_check, k))
         if re.match(regexp, str(value_to_check)):
             continue
         else:
             raise e.InvalidRequestField(k)
```
bassosimone: (here you don't need the else because there is continue above)

bassosimone:
@hellais reading this diff, I understand that `PUT /report` is used to update a report, but in TheTorProject/ooni-probe#456 it seems to me that `POST /report/$id` is used.

oonib/report/handlers.py
hellais: yeah you are right. I suspect this may be due to some rebase mess up. Anyways very good catch.

oonib/testhelpers/http_helpers.py
hellais: Yeah you are right this is a leftover from the cherry-ditching of the web_connectivity test helper.

oonib/report/handlers.py
hellais: Because parseUpdateReportRequest takes an optional report_id parameter that allows me to inject another report ID in there if it's not already present.

In the line `report_id = parsed_request.get('report_id', report_id)` I get the report_id and if it's not present I set it to the argument passed to the fore-mentioned function.

oonib/test/test_report.py
hellais: True. I will implement a test case for this.

oonib/report/handlers.py
bassosimone: ack!

oonib/report/handlers.py
bassosimone: True, thanks for explaining!

bassosimone:
Replying to my previous comment: actually the code allows to both `POST /report/$id` and `PUT /report`. The former is implemented by `UpdateReportHandlerFile`. Explained by @hellais in chat.

bassosimone:
That's what needs to be done before we could merge this pull request:

- [x] review the pull request
- [ ] test the code

bassosimone:
:+1: 

Okay, I have tested this. I run `oonib` using Vagrant `precise32` box. In another shell I run `ooniprobe` using a deck and configuring the custom `oonib` as collector. I stopped the deck after about one hour. Then I verified that all the json lines inside the report could be loaded using `json.loads()`.

I think this is enough to declare this PR mergeable and proceed with merging!

PR #63, opened 2016-03-08
Feature/web connectivity

hellais:


oonib/testhelpers/http_helpers.py
```diff
@@ -1,16 +1,87 @@
...
+from cyclone.web import RequestHandler, Application, HTTPError
+from cyclone.web import asynchronous
 
 from twisted.protocols import policies, basic
 from twisted.web.http import Request
 
 from oonib import log, randomStr
 
+# Monkey patch to overcome issue in parsing of HTTP responses that don't
+# contain 3 parts in the HTTP response line
```
bassosimone: Curiosity: in which case do this happen?

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +239,282 @@ def all(self, length, keyword):
...
+
+        expiration.reset(self.expiration_time)
+        lock.release()
+        defer.returnValue(value)
+
+    @defer.inlineCallbacks
+    def http_request(self, url):
+        cached_value = yield self.lookup('http_request', url)
+        if cached_value is not None:
+            defer.returnValue(cached_value)
```
bassosimone: Remind me: so this `returnValue` mocks a proper return by throwing an exception, right?

oonib/testhelpers/http_helpers.py
```diff
@@ -1,16 +1,87 @@
...
+def statusReceivedPatched(self, status):
+    try:
+        return old_status_received(self, status)
+    except ParseError as exc:
+        if exc.args[0] == 'wrong number of parts':
+            return old_status_received(self, status + " XXX")
+        raise
+HTTPClientParser.statusReceived = statusReceivedPatched
+
+class FixedRedirectAgent(BrowserLikeRedirectAgent):
```
bassosimone: Interesting to see that the addition of web connectivity and the addition of logic on the server side lead to (increased?) code sharing between ooni-probe and ooni-backend!

bassosimone:
- [ ] first reading of the diff
- [ ] second reading of the diff
- [ ] testing

oonib/testhelpers/http_helpers.py
hellais: It happens because some people decided to write their own HTTP server that doesn't fully respect the HTTP spec. An example of this okcupid.com:

```
curl -kis okcupid.com

HTTP/1.1 302
Date: Tue, 24 May 2016 10:40:50 GMT
Transfer-Encoding: chunked
Connection: keep-alive
Set-Cookie: __cfduid=dda563ef0b63790ff641999bbb5432a231464086450; expires=Wed, 24-May-17 10:40:50 GMT; path=/; domain=.okcupid.com; HttpOnly
X-OKWS-Version: OKWS/3.1.38.0
```

What's weird is that from the source code they have published it looks like it actually does support setting the status code reason string (https://github.com/okws/okws/blob/80f41cc0ff770a83e9f9e468cfdc2040dd7e0c43/libahttp/resp.h#L55), but in their deployment it doesn't. Probably this is due to some other middle box mangling with the response line.

oonib/testhelpers/http_helpers.py
hellais: Yeah, I am actually a bit scared by this tbh since it's not kept in sync and bugs found in ooniprobe or oonibackend need to be manually back ported to the other.

In the future maybe it will make sense to have a common library shared by both.

oonib/testhelpers/http_helpers.py
hellais: Yeah, IRC what it does is it raises a `StopIteration` exception so that the generator is interrupted and the value passed to return value is passed as a callback argument to the deferred.

oonib/testhelpers/http_helpers.py
bassosimone: [In a recent RFC](https://tools.ietf.org/html/rfc7230#section-3.1.2) the grammar for the response line is:

```
status-line = HTTP-version SP status-code SP reason-phrase CRLF
reason-phrase  = *( HTAB / SP / VCHAR / obs-text )
```

Unless the `*` has a very surprising meaning, I'd say it would be theoretically possible to omit the reason phrase, but I'd also say an implementation SHOULD NOT

bassosimone: Also, [if I read correctly this part of nodejs/http-parser](https://github.com/nodejs/http-parser/blob/master/http_parser.c#L888), it seems that nodejs/http-parser copes with a missing status code allowing CR to appear just after a number in the status line.

oonib/testhelpers/http_helpers.py
bassosimone: That's very twisted of Twisted!

bassosimone:
Will now do another pass of reading the diff and later will follow up with testing of this branch. I am going to give priority to the backend and then move on the test the frontend.

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +241,336 @@ def all(self, length, keyword):
             length = 100000
         self.write(self.genRandomPage(length, keyword))
 
+
+def encodeResponse(response):
+    body = None
+    body_length = 0
+    if getattr(response, 'body', None):
```
bassosimone: Wouldn't be simpler written as `hasattr(response, 'body')`?

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +241,336 @@ def all(self, length, keyword):
...
+        if name.lower() not in self._rawHeaders:
+            self._rawHeaders[name.lower()] = dict()
+        self._rawHeaders[name.lower()]['name'] = name
+        self._rawHeaders[name.lower()]['values'] = values
+
+    def getAllRawHeaders(self):
+        for k, v in self._rawHeaders.iteritems():
+            yield v['name'], v['values']
+
+    def getRawHeaders(self, name, default=None):
```
bassosimone: The name of this method: shouldn't it be `getRawHeader`?

hellais: This is mapped directly to the interface of the `twisted.web.http_headers.Headers` class. It's in the plural because actually there can be multiple headers for a given header name.

oonib/testhelpers/http_helpers.py
hellais: Actually, yes 👍 

Will fix.

oonib/testhelpers/http_helpers.py
hellais: Yes, I think we should make a module called something like `txextra` and put all this stuff in there.

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +241,336 @@ def all(self, length, keyword):
...
+            'tcp_connect': tcp_connect,
+            'dns': dns
+        })
+
+    @asynchronous
+    def post(self):
+        try:
+            request = json.loads(self.request.body)
+            self.control_measurement(
+                str(request['http_request']),
```
bassosimone: Why here there is need to cast to `str()` explicitly?

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +241,336 @@ def all(self, length, keyword):
...
+        http_request = responses[0][1]
+        dns = responses[1][1]
+        tcp_connect = {}
+        for idx, response in enumerate(responses[2:]):
+            tcp_connect[socket_list[idx]] = response[1]
+        self.finish({
+            'http_request': http_request,
+            'tcp_connect': tcp_connect,
+            'dns': dns
+        })
```
bassosimone: Running the backend in a local Vagrant and using `curl(1)` as follows:

```
curl -vX POST --data @INPUT http://127.0.0.1:57007/
```

where `INPUT` contains

```
{"http_request": "https://ooni.torproject.org/","tcp_connect":["154.35.132.70","82.195.75.101","38.229.72.16","138.201.14.197","86.59.30.40"]}
```

it yields:

```
* About to connect() to 127.0.0.1 port 57007 (#0)
*   Trying 127.0.0.1... connected
> POST / HTTP/1.1
> User-Agent: curl/7.22.0 (i686-pc-linux-gnu) libcurl/7.22.0 OpenSSL/1.0.1 zlib/1.2.3.4 libidn/1.23 librtmp/2.3
> Host: 127.0.0.1:57007
> Accept: */*
> Content-Length: 142
> Content-Type: application/x-www-form-urlencoded
> 
* upload completely sent off: 142out of 142 bytes
```

and then `curl(1)` is stuck more or less forever (depending on how you define "forever").

Now, looking at the `oonib` side, one sees this in the logs:

```
--- <exception caught here> ---
  File "/usr/local/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1128, in _inlineCallbacks
    result = g.send(result)
  File "/data/oonib/oonib/testhelpers/http_helpers.py", line 478, in tcp_connect
    ip_address, port = socket.split(":")
exceptions.ValueError: need more than 1 value to unpack
```

This error is caused by the fact that I supplied wrong input (the port was not specified). But what is interesting is that no error is ever returned to the client.

I believe this happens because `@synchronous` tells cyclone that `finish` must explicitly be called whilean exception is raised when calling `web_connectivity_cache.tcp_connect` and hence the final `finish` function is never called.

If I'm right, the proper fix would probably be to wrap the body of `control_measurement` with `try..catch` and to make sure that `self.finish()` is called also on exception -- or, possibly, as part of `finally`.

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +241,336 @@ def all(self, length, keyword):
...
+                        raise
+                    retries += 1
+        except DNSLookupError:
+            page_info['failure'] = 'dns_lookup_error'
+        except TimeoutError:
+            page_info['failure'] = 'generic_timeout_error'
+        except ConnectionRefusedError:
+            page_info['failure'] = 'connection_refused_error'
+        except:
+            # XXX map more failures
```
bassosimone: What other failures shall be mapped here? Shall we need to create an issue to remind us that?

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +241,336 @@ def all(self, length, keyword):
...
+                    if self.enable_debug:
+                        response.body = body
+                        page_info['responses'] = encodeResponses(response)
+                    break
+                except:
+                    if retries > self.http_retries:
+                        raise
+                    retries += 1
+        except DNSLookupError:
+            page_info['failure'] = 'dns_lookup_error'
```
bassosimone: wishful thinking: it would be nice to have a `dict` mapping classes to their ooni error string counterparts

oonib/testhelpers/http_helpers.py
bassosimone: I see!

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +241,336 @@ def all(self, length, keyword):
...
+                    response = yield agent.request('GET', url,
+                                                   TrueHeaders(REQUEST_HEADERS))
+                    headers = {}
+                    for name, value in response.headers.getAllRawHeaders():
+                        headers[name] = unicode(value[0], errors='ignore')
+                    body_length = -1
+                    try:
+                        body = yield readBody(response)
+                        body_length = len(body)
+                    except PartialDownloadError as pde:
```
bassosimone: I do not understand if this error is also triggered when the "until EOF" body semantic is used or whether this case only catches when the connection is unexpectedly interrupted, but I suspect it's the former.

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +241,336 @@ def all(self, length, keyword):
...
+                        if pde.response:
+                            body_length = len(pde.response)
+                    page_info['body_length'] = body_length
+                    page_info['status_code'] = response.code
+                    page_info['headers'] = headers
+                    page_info['title'] = extractTitle(body)
+                    if self.enable_debug:
+                        response.body = body
+                        page_info['responses'] = encodeResponses(response)
```
bassosimone: I think I missed so far whether this option of enabling debug could be enabled by client action or is a server side only property. I see cases whether it would be useful to ask the backend to retrieve more. Thoughts?

bassosimone:
Possibly unrelated, but: while testing this branch I've started to notice this txtorcon error (adding a bit of context before the actual error):

```
54%: Loading relay descriptors
56%: Loading relay descriptors
59%: Loading relay descriptors
63%: Loading relay descriptors
69%: Loading relay descriptors
75%: Loading relay descriptors
80%: Connecting to the Tor network
90%: Establishing a Tor circuit
100%: Done
Application starting on 42738
Unhandled Error
Traceback (most recent call last):
Failure: txtorcon.torcontrolprotocol.TorProtocolError: 513 Unacceptable option value: Failed to configure rendezvous options. See logs for details.
```

@hellais, any hint at what this error means? That's what I changed from `master`:

```sh
$ diff -u oonib.conf{.example,}
--- oonib.conf.example	2016-05-26 17:56:33.000000000 +0000
+++ oonib.conf	2016-05-26 19:49:04.000000000 +0000
@@ -75,3 +75,4 @@
     web_connectivity:
         endpoints:
         - {type: tcp, port: 57007}
+        - {type: tls, port: 57008, cert: "private/ssl-key-and-cert.pem"}
$ git diff
diff --git a/data/bouncer.yaml b/data/bouncer.yaml
index 6ef6b71..9d5feb8 100644
--- a/data/bouncer.yaml
+++ b/data/bouncer.yaml
@@ -17,3 +17,5 @@ collector:
       - {name: multiprotocol_traceroute_test, version: 0.1.1}
     test-helper: {dns: '93.95.227.200:57004', http-return-json-headers: 'http://93.95.227.200',
       ssl: 'https://93.95.227.20', tcp-echo: 213.138.109.232, traceroute: 93.95.227.200}
+    test-helper-alternate: {dns: '13.95.227.200:57004', http-return-json-headers: 'http://13.95.227.200',
+      ssl: 'https://13.95.227.20', tcp-echo: 13.138.109.232, traceroute: 13.95.227.200}
```

oonib/testhelpers/http_helpers.py
hellais: Because the result of `json.loads` is `unicode` and twisted wants to have a `str()` as a URL.

oonib/testhelpers/http_helpers.py
hellais: Good call.

As suggested I think it's best to convert the `finish()` above into a `write()` and call the finish inside of the finally below.

oonib/testhelpers/http_helpers.py
hellais: Yes actually I think this would be a very useful feature to turn on and off. I am going to add a boolean key to the request of the client called `include_http_responses` that will allow to return this value.

I guess this means that we would have to increase the size of the cache, but since it's being stored on disk it shouldn't be a bit issue.

oonib/testhelpers/http_helpers.py
hellais: Yes that is a good idea. Will make a ticket for it.

oonib/testhelpers/http_helpers.py
hellais: What do you mean by "until EOF"?

Generally it will happen when the connection is unexpectedly interrupted.

hellais:
@bassosimone that error appears to be related to sending an invalid command via the Tor control port (see: https://gitweb.torproject.org/torspec.git/tree/control-spec.txt#n1534).

Do you have the tor log for when that error occurs? Does this happen every time you run it or is it not reproducible every time?

Also note that the correct format for `test-helper-alternate` is to specify them as lists like so:

```
test-helpers-alternate: {
     "web-connectivity": [
              {'type': 'https | onion | http', 'address': 'http://foo.com'},
              ... 
     ]
}
```

oonib/testhelpers/http_helpers.py
bassosimone: Thanks! For the records it's issue #81.

oonib/testhelpers/http_helpers.py
bassosimone: I mean that [the seventh -- in order of importance -- possible ways to indicate body length is](https://tools.ietf.org/html/rfc7230#section-3.3.3):

> 7.  Otherwise, this is a response message without a declared message
>      body length, so the message body length is determined by the
>      number of octets received prior to the server closing the
>      connection.

This way of indicating body length is not optimal, because obviously one cannot tell whether the server rightfully closed connection or whether, say, an error occurred forcing shutdown of the child serving the request. Yet, it is used in the wild in many cases.

So, my question was whether this Twisted error is used to indicate that specific use case of HTTP or whether Twisted had another way of reporting this case. Was asking this because, depending on the response, I am not sure to consider it correct to consider it good enough to use it.

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +190,367 @@ def all(self, length, keyword):
...
+                        body = yield readBody(response)
+                        body_length = len(body)
+                    except PartialDownloadError as pde:
+                        if pde.response:
+                            body_length = len(pde.response)
+                    page_info['body_length'] = body_length
+                    page_info['status_code'] = response.code
+                    page_info['headers'] = headers
+                    page_info['title'] = extractTitle(body)
+                    response.body = body
```
bassosimone: Also related to the `PartialDownloadError` observation above: in case such exception is raised, what is the value of the `body` variable? I think there could be cases where the variable does not exist, not sure instead if there is a case where it could be the body of the previous loop cycle, if the previous loop cycle does not end well.

So, assuming that we *want* to use the partial body, I guess that the branch catching the partial download exception shall make sure that variables required by the code filling `page_info` are always defined.

oonib.conf.example
```diff
@@ -30,12 +30,13 @@ main:
     tor_datadir: null
 
     bouncer_endpoints:
-    - {type: tls, port: 10443, cert: "private/ssl-key-and-cert.pem"}
+    #- {type: tls, port: 10443, cert: "private/ssl-key-and-cert.pem"}
     - {type: tcp, port: 10080}
-    - {type: onion, hsdir: bouncer}
+    - {type: onion, hsdir: /tmp/bouncer}
```
bassosimone: Question: so, I understand that `/tmp/bouncer` should fix the error that I reported, right? Why was the `tls` entry commented out? If I recall correctly, when firstly testing, I received an exception because certificate was not there and then I did generate the certificate using the correct script, and then everything worked correctly AFAICT. If possible, thus, I'd leave the `tls` line not commented, such that by default one tests as much features as possible of oonib.

oonib/txextra.py
```diff
@@ -0,0 +1,81 @@
+from twisted.web.http_headers import Headers
```
bassosimone: Curiosity: what is the meaning of `tx` in this context?

oonib/txextra.py
```diff
@@ -0,0 +1,81 @@
...
+            raise ResponseFailed([Failure(err)], response)
+        location = self._resolveLocation(response.request.absoluteURI, locationHeaders[0])
+        deferred = self._agent.request(method, location, headers)
+
+        def _chainResponse(newResponse):
+            newResponse.setPreviousResponse(response)
+            return newResponse
+
+        deferred.addCallback(_chainResponse)
+        # This is the fix to properly handle redirects
```
bassosimone: I do not understand the comment above. AFAICT the fix to properly handle redirects is using `response.request.absoluteURI` on line 44.

oonib/txextra.py
bassosimone: Ah, perhaps, Twisted extension?

oonib/txextra.py
```diff
@@ -0,0 +1,81 @@
...
+        return deferred.addCallback(
+            self._handleResponse, method, uri, headers, redirectCount + 1)
+
+class TrueHeaders(Headers):
+    def __init__(self, rawHeaders=None):
+        self._rawHeaders = dict()
+        if rawHeaders is not None:
+            for name, values in rawHeaders.iteritems():
+                if type(values) is list:
+                    self.setRawHeaders(name, values[:])
```
bassosimone: I guess that `values[:]` is some Pythonic trick here, right? You are basically taking all elements in the list, right?

oonib/txextra.py
```diff
@@ -0,0 +1,81 @@
...
+                    self.setRawHeaders(name, values)
+
+    def setRawHeaders(self, name, values):
+        if name.lower() not in self._rawHeaders:
+            self._rawHeaders[name.lower()] = dict()
+        self._rawHeaders[name.lower()]['name'] = name
+        self._rawHeaders[name.lower()]['values'] = values
+
+    def getAllRawHeaders(self):
+        for k, v in self._rawHeaders.iteritems():
```
bassosimone: Here you can probably write `for _, v in ...`

oonib/txextra.py
```diff
@@ -0,0 +1,81 @@
...
+
+class TrueHeaders(Headers):
+    def __init__(self, rawHeaders=None):
+        self._rawHeaders = dict()
+        if rawHeaders is not None:
+            for name, values in rawHeaders.iteritems():
+                if type(values) is list:
+                    self.setRawHeaders(name, values[:])
+                elif type(values) is dict:
+                    self._rawHeaders[name.lower()] = values
```
bassosimone: Perhaps stupid question because I don't understand all the context, but: so, here you're setting the `dict` named `values` as value and below you assume that such `dict` would contain the `name` and `values` keys, but here you don't make sure it contains such keys before setting it... is this something that could create issues or antani?

oonib/testhelpers/http_helpers.py
```diff
@@ -242,21 +191,49 @@ def all(self, length, keyword):
         self.write(self.genRandomPage(length, keyword))
 
 
+META_CHARSET_REGEXP = re.compile('<meta(?!\s*(?:name|value)\s*=)[^>]*?charset\s*=[\s"\']*([^\s"\'/>]*)')
+
+def representBody(body):
```
bassosimone: Going on memory: isn't this function also shared with ooni-probe?

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +190,367 @@ def all(self, length, keyword):
             length = 100000
         self.write(self.genRandomPage(length, keyword))
 
+
+META_CHARSET_REGEXP = re.compile('<meta(?!\s*(?:name|value)\s*=)[^>]*?charset\s*=[\s"\']*([^\s"\'/>]*)')
+
+def representBody(body):
+    # XXX perhaps add support for decoding gzip in the future.
```
bassosimone: Going on memory: isn't `representBody` also shared with ooni-probe?

oonib/testhelpers/http_helpers.py
```diff
@@ -168,7 +190,367 @@ def all(self, length, keyword):
...
+            self._cache_lifecycle[response_type][key_hash]['expiration']
+
+        yield lock.acquire()
+
+        if not os.path.exists(cache_file):
+            lock.release()
+            defer.returnValue(None)
+
+        with open(cache_file, 'r') as fh:
+            value = json.load(fh)
```
bassosimone: Wow: I really like the cache code: it is very beautiful!

bassosimone:
Cool: I can confirm that 10e3ce7b4c5bf fixes the txtorcon error I reported, thanks!

oonib/testhelpers/http_helpers.py
hellais: yes this is true.

oonib/testhelpers/http_helpers.py
bassosimone: I can confirm that passing invalid input now yields an HTTP error after d741df28a0c7b61a!

oonib/txextra.py
hellais: Yes you are right this comment is misplaced. The actual fix is to consider the whole URI (https://twistedmatrix.com/trac/attachment/ticket/8265/fix-redirect-resolve-location.patch)

oonib/txextra.py
hellais: tx is a convention used to indicate third-party "Twisted" extensions, for example `txtorcon`, `txsocksx`. I think the origin is "Twisted eXtension".

bassosimone:
Two further considerations regarding the caching algorithm:

1) It is my understanding that [this line of code](https://github.com/TheTorProject/ooni-backend/pull/63/files#diff-c866ac5417d994fb5427137aa2b2b4f4R356) *unconditionally* resets the expire time of the cache entry. I am bit troubled by the unconditionally part of it. 

In theory, if there are enough requests for the same resource to the backend, one entry could never expire, correct?

I am not sure this behavior is correct in general, and I'd rather *unconditionally* fetch again the resource after some time. I feel like that would increase robustness and scalability of the backend.

2) I tested on the Vagrant system I am using with the following input:

```
$ cat IN4
{
  "http_request": "https://popcorntime.io/",
  "tcp_connect": [
    "52.70.175.181:443"
  ]
}
```

and using this curl command:

```
curl -vXPOST -H 'Content-Type: application/json' --data @IN4 \
    http://127.0.0.1:57007/ | python -m json.tool
```

Now, since popcorntime.io IP address is blocked in Italy and since timeouts seems to be very large, this request takes more than two minutes to complete on my system.

What troubles me is the following scenario: a slow request to the backend is pending (such as, in my case, the one in `IN4`) and another request for the same resource arrives. In such case, another slow request is started, because no entry is added to the cache, resulting in *a second* slow operation to be scheduled by the backend and in the cache being overwritten when the request completes.

I am not sure this behavior is optimal in terms (again) of robustness and scalability, because it leads to higher load on the backend. It probably would be better in terms of scalability if the first request triggered the backend to do the DNS, tcp-connect, and HTTP dance, and then second request only registered another observer for the same input, and when the (let's call it like this) background thread doing the fetching completes, both pending requests are notified. Let me note, *en passant*, that this seems a thing that it would be easy to implement in Go using messages and coroutines.

Speaking of moving forward, I suggest to create issues for these two items highlighted above (provided that there is agreement that they *are* issues) and to not attempt to fix them as part of this pull request. I think what we have here is a very good first iteration at solving the smarter backend problem and that piling up further changes on top of this branch would only further complicate the current review process and possible future attempts at understanding what happened and bisecting.

This is to say: as far as I am concerned, the review process is complete and, once other minor comments have been addressed and these two issues ticketized, I am all for merging this branch.

oonib.conf.example
hellais: I commented it out because I am thinking that it would be good that by default when you start oonib using vagrant it "just works" instead of having to run some extra scripts.

Since it's in the example people should be able to figure out how to make it work, but I wouldn't want to degrade the experience with unattended setup.

bassosimone: Okay, the "just works" argument stands. It touches some OpenBSD chords inside me :-).

oonib/txextra.py
hellais: Yes this takes all elements in the list, but also creates a copy in the process of doing so. I think the reason why we do this is that otherwise we would be pointing to a list inside of another rawHeaders object and hence changing values inside that other object would lead to them being changed in here too.

bassosimone: These Pythonic tricks!

oonib/txextra.py
bassosimone: I see, thanks!

oonib/txextra.py
bassosimone: ack, thanks!

oonib/testhelpers/http_helpers.py
bassosimone: So, it would probably make sense to move this piece of code in `txextra`. I think it would be okay to do this refactoring later and for now only add a ticket for it.

oonib/txextra.py
hellais: To be honest I am not seeing this way of using the constructor at all in the codebase. I am thinking that it's perhaps ideal to just remove this way of constructing the TrueHeaders object as you are right it does add a bit of complexity to it.

oonib/txextra.py
hellais: +1

oonib/testhelpers/http_helpers.py
hellais: Yes this is a good point.

I will adjust this accordingly.

hellais:
These are two very good points.

On the topic of 1) I think you are right and we should probably be setting an upper bound on how much a certain resource can stay in the cache and have an upper bound expiration time on every resource (otherwise we could potentially end up serving indefinitely stale content).

Regarding 2) I think it's not too complex to implement what you suggest. Probably what we want to be doing is acquire the lock on the cache as soon as we begin doing any resource lookup so that the following requests will be waiting on the lock for the cache lookup and will get the cached value as soon as the lock is released (the resource has been cached).

As you said I think these are two tricky bits that are probably best deferred to being implemented in a second iteration.

I am going to file some tickets for them and proceed in merging this.

bassosimone:
> Probably what we want to be doing is acquire the lock on the cache as soon as we begin doing any resource lookup

Ah, right, that would be a very good way to move forward!

oonib/testhelpers/http_helpers.py
bassosimone: Thanks!

bassosimone:
With @hellais we have further investigated the `PossibleDataLoss` issue that I raised earlier and concluded that the current code is fine, because `PossibleDataLoss` is Twisted's way of telling the user that no content-length was provided and hence Twisted cannot guarantee all the body was downloaded.

PR #67, opened 2016-03-17
Change the value of the key that is sent by the client in the 0.2 dat…

hellais:
…aformat


bassosimone:
I have tested this diff both with master and with TheTorProject/ooni-probe#456 and verified that actual files were written in the `data/archive` directory. I have not run all tests but a reasonable subset of them to declare, in my opinion, safe to merge this branch into master. :+1:

PR #68, opened 2016-03-30
Set the report file template in the example configuration to the default

hellais:


bassosimone:
Tested, it works! Thank you!

PR #69, opened 2016-04-24
[WIP] Add TLS endpoint

joelanders:
Ok, I think this is all working.
Be careful to hit the endpoint with an explicit `https`, eg: `https://localhost:whatever` as I wasted a bit of time on this :p


oonib.conf.example
```diff
@@ -28,7 +27,17 @@ main:
...
+    tor_datadir: test_datadir
+
+    bouncer_endpoints:
+    - {type: tls, port: 10443, cert: "private/ssl-key-and-cert.pem"}
+    - {type: tcp, port: 10080}
+    - {type: onion, hsdir: bouncer}
+
+    collector_endpoints:
+    - {type: tls, port: 11443, cert: "private/ssl-key-and-cert.pem"}
+
```
hellais: It would be ideal if we could keep backward compatibility with old configuration files.

I am thinking that instead of having the endpoints be a list we could expand them to something like:
```
bouncer_tls_port: XXX
bouncer_tls_cert: XXX
bouncer_hsdir: XXX

collector_tls_port: XXX
collector_tls_cert: XXX
etc.
```

oonib/oonibackend.py
```diff
@@ -102,60 +102,71 @@
...
+                                        data_dir=hsdir)
+
+def getTCPEndpoint(endpoint_config):
+    return endpoints.TCP4ServerEndpoint(reactor, endpoint_config['port'])
+
+def getTLSEndpoint(endpoint_config):
+    with open(endpoint_config['cert'], 'r') as f:
+        cert_data = f.read()
+    certificate = ssl.PrivateCertificate.loadPEM(cert_data)
+    print certificate.inspect()
```
hellais: You probably want to remove this print statement.

oonib/oonibackend.py
```diff
@@ -102,60 +102,71 @@
     multiService.addService(http_return_request_helper)
     http_return_request_helper.startService()
 
-# add the tor collector service here
+def getHSEndpoint(endpoint_config):
+    hsdir = os.path.join(torconfig.DataDirectory, endpoint_config['hsdir'])
```
hellais: The directory should be set by default to either "collector" or "bouncer" to maintain backward compatibility.

Moreover I would expect that if I set the `hsdir` directory to an absolute path it would end up there and not be expanded to a subdirectory of the tor data directory.

oonib/oonibackend.py
```diff
@@ -102,60 +102,71 @@
...
-                                            80,
-                                            hidden_service_dir=data_dir)
-        else:
-            return TCPHiddenServiceEndpoint(reactor,
-                                            torconfig,
-                                            80,
-                                            data_dir=data_dir)
-
-    def printOnionEndpoint(endpointService):
-        print ("Exposed %s Tor hidden service on httpo://%s" %
```
hellais: We should maintain the ability to be able to see on which address the HS is exposed on.

oonib/oonibackend.py
```diff
@@ -102,60 +102,71 @@
...
+    if endpoint_config['type'] == 'onion':
+        return getHSEndpoint(endpoint_config)
+    elif endpoint_config['type'] == 'tcp':
+        return getTCPEndpoint(endpoint_config)
+    elif endpoint_config['type'] == 'tls':
+        return getTLSEndpoint(endpoint_config)
+    else:
+        raise Exception("unknown endpoint type")
+
+class Echo(protocol.Protocol):
```
hellais: This class should probably be removed.

oonib/oonibackend.py
```diff
@@ -102,60 +102,71 @@
...
-    if ooniBouncer:
-        bouncer_service = internet.TCPServer(8888, ooniBouncer,
-                                             interface="127.0.0.1")
-        multiService.addService(bouncer_service)
-        bouncer_service.startService()
-    collector_service = internet.TCPServer(8889, ooniBackend,
-                                           interface="127.0.0.1")
-    multiService.addService(collector_service)
-    collector_service.startService()
+    configTor(torconfig)
```
hellais: From the looks of it tor is no longer being started. How are the hidden services being brought up without tor?

hellais:
Overall this looks good 👍  I made a bit of comments for things that should be changed before we can merge this. Good stuff.

oonib.conf.example
joelanders: I kinda thought it was nicer to group all the needed things into an object. It'd be even nicer if I made that "object" the Twister ServerFromString or whatever.

oonib/oonibackend.py
joelanders: yes

oonib/oonibackend.py
joelanders: There's also a `tor_datadir` config variable, and I thought it'd be more confusing if one took priority (sometimes). Maybe erroring out is the least confusing?

oonib/oonibackend.py
joelanders: Yeah, it looks like it doesn't clearly say "bouncer: sdflh49hd.onion"

oonib/oonibackend.py
joelanders: whoops

oonib/oonibackend.py
joelanders: txtorcon's `TCPHiddenServiceEndpoint.global_tor()` starts a global tor the first time it's called, then uses the existing global tor when it's called again.

oonib/oonibackend.py
joelanders: current log:
```
Started hidden service "ecvvd5yhektgxm3i.onion" on port 80
Keys are in "/home/landers/code/Tor/ooni-backend/test_datadir/collector".
100% At least one descriptor uploaded.
Started hidden service "wvcw5wzy7f3uwoml.onion" on port 80
Keys are in "/home/landers/code/Tor/ooni-backend/test_datadir/bouncer".
```

joelanders: Balls, I'm creating a `StreamServerEndpointService` from a `factory` and an `endpoint` and I can't find (in the few minutes I have right now) where to add a callback.

If you needed the TLS bit for something this week, I can figure out the logging later.

oonib/oonibackend.py
hellais: Yes you are right, should have checked that.

This is actually a very good improvement as I believe it also fixes the issues we were having with priviledge shedding and Tor not starting shown here: https://github.com/TheTorProject/ooni-backend/issues/65

gen-ssl-key-cert.sh
```diff
@@ -0,0 +1,13 @@
...
+if [ ! -d private ]
+then
+  mkdir private
+fi
+
+openssl req -x509 -newkey rsa:2048 \
+    -keyout private/ssl-key.pem -out private/ssl-cert.pem \
+    -days 400 -nodes -subj '/CN=selfie'
+
+cat private/ssl-key.pem private/ssh-cert.pem > private/ssl-key-and-cert.pem
```
hellais: I am going to move this into the scripts directory when merging.

oonib.conf.example
hellais: Ok, let's keep this as is for the moment. I noticed you also added backward compatible changes so I think it's ok for the moment.

oonib/oonibackend.py
hellais: The `tor_datadir` is actually just the directory where tor is instructed to store it's various files, but it's not really a requirement, nor is it how usually tor is configured, for the hidden service data files to reside in the tor data directory.

I would say we could have a policy of these sorts:

* If `hsdir` is set, then it should be considered as a path that is not relative to `tor_datadir`

* If `hsdir` is not set, then we follow the legacy behaviour of setting the collector `hsdir` to `$tor_datadir/collector` and the bouncer `hsdir` to `$tor_datadir/bouncer`.

hellais:
Other things to take into consideration:

* [ ] If `config.main.tor_hidden_service` is set to false, then calls to `getHSEndpoint` will fail due to `torconfig` not being defined

* [ ] Make sure the documentation is updated to reflect the changes in the configuration file

* [ ] Update ooni-sysadmin to reflect the changes in this repo

hellais:
I created two tickets for the documentation/sysadmin tasks here:

* https://github.com/TheTorProject/ooni-sysadmin/issues/49

* https://github.com/TheTorProject/ooni-backend/issues/72

I am going to to a final review of this branch and merge it.

hellais:
Now there is an issue with the fact that if you set `tor_hidden_service: false` and don't set the `bouncer_endpoints` or `collector_endpoints` to be a list (because for example you want to just run test helpers and not have a bouncer or a collector) it will try to iterate over a None-type:

```
    for endpoint_config in config.main.bouncer_endpoints:
TypeError: 'NoneType' object is not iterable
```

hellais:
I fixed the outstanding bugs and merged into master.

Thanks!

Issue #73, opened 2016-05-13
New HS endpoint code doesn't respect tor_datadir


joelanders:
I introduced an embarrassing regression in PR #69 where the `torconfig` isn't passed to the endpoint-starting code. It's now taking the default config from txtorcon and using a temporary datadir.

joelanders:
https://github.com/meejah/txtorcon/blob/master/txtorcon/endpoints.py#L215
`TCPHiddenServiceEndpoint.global_tor()` doesn't take a `tor_datadir` argument.

Maybe we should take the advice (a few lines down in the above link)?:
```
It's probably best to not specify any option besides
`public_port`, `hidden_service_dir`, and maybe `local_port`
unless you have a specific need to.
```

Do we have a specific need for
* socks_port (we're never using it here)
* control_port (we can let txtorcon choose this)
* tor2web mode
* tor_datadir (txtorcon generates a safe tempdir, do we ever need to persist the contents?)

(We could get rid of all of `oonib/onion.py` if we don't need these things)

hellais:
I think this can be closed now.

PR #74, opened 2016-05-14
Simplify config

joelanders:
https://github.com/TheTorProject/ooni-backend/issues/73

As mentioned in #73, I accidentally changed the tor_data_dir behavior.

This is what the code might look like if we just take the txtorcon defaults for things.

(I need to double check this code again for backwards compatibility and handling of unset values.)


oonib/onion.py
```diff
@@ -1,91 +0,0 @@
...
-def configTor(torconfig):
-    def updates(prog, tag, summary):
-        print("%d%%: %s" % (prog, summary))
-
-    if config.main.socks_port:
-        torconfig.SocksPort = config.main.socks_port
-    if config.main.control_port:
-        torconfig.ControlPort = config.main.control_port
-    if config.main.tor2webmode:
-        torconfig.Tor2webMode = 1
```
hellais: We actually need to set this in order to be able to run our custom compiled version of Tor with the tor2web mode enabled.

oonib.conf.example
```diff
@@ -7,11 +7,6 @@ main:
     bouncer_file: data/bouncer.yaml
 
     logfile: null
-    database_uri: 'sqlite://oonib_test_db.db'
-    db_threadpool_size: 10
-    tor_binary: null
```
hellais: We actually need the ability to specify a custom tor binary so we can use tor2web mode. I actually didn't realise that this had already been removed in the previous PR, but it should go back in.

PR #75, opened 2016-05-15
keep old config behavior with new endpoint code

joelanders:
https://github.com/TheTorProject/ooni-backend/issues/73

Here's another attempt at maintaining the old behavior with the new endpoint code.

I could remove the inlineCallbacks from getHSEndpoint and createService if we allow it to be an error to omit tor_datadir when using an old style configuration (a config without the *_endpoints lines).

I have to complicate those functions here because we don't know the hsdir until after tempfile.mkdtemp()... bleh

### DIFF
--- a/oonib/onion.py
+++ b/oonib/onion.py
@@ -1,7 +1,7 @@
 import tempfile
 from oonib import log
 from oonib.config import config
-from twisted.internet import reactor, endpoints
+from twisted.internet import reactor, endpoints, defer
 import os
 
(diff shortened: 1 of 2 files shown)


joelanders:
Bleh, I'm looking at this again and it seems more complicated than it needs to be. I can give it another go.

hellais:
If you have something in mind on how to simplify this that would be great.

I will let you finish that then 👍

PR #77, opened 2016-05-21
keep old config behavior with new endpoint code

joelanders:
Ok, to recap: my TLS endpoint PR used txtorcon's `TCPHiddenServiceEndpoint.global_tor()` which takes care of starting and maintaining a reference to a singleton tor instance. Unfortunately, the configuration values are hardwired, and we need some options not in the defaults (pointing at a tor2web-enabled tor binary out of the $PATH, for example).

So in this PR, I write a `get_global_tor()` function largely the same as what txtorcon has, but using our existing configuration code instead of the txtorcon defaults. Then I replace `TCPHiddenServiceEndpoint.global_tor()` with `TCPHiddenServiceEndpoint()` and `get_global_tor()`.

Also, the old configuration file format should produce the same behavior with the new code.

### DIFF
--- a/oonib/oonibackend.py
+++ b/oonib/oonibackend.py
@@ -11,14 +11,14 @@ from distutils.version import LooseVersion
 
 from oonib.api import ooniBackend, ooniBouncer
 from oonib.config import config
-from oonib.onion import configTor
+from oonib.onion import get_global_tor
 from oonib.testhelpers import dns_helpers, ssl_helpers
 from oonib.testhelpers import http_helpers, tcp_helpers
 
 import os
 
 from twisted.application import internet, service
-from twisted.internet import reactor, endpoints, ssl
+from twisted.internet import reactor, endpoints, ssl, defer
 from twisted.names import dns
 
 from txtorcon import TCPHiddenServiceEndpoint, TorConfig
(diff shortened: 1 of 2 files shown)


oonib/oonibackend.py
```diff
@@ -151,27 +153,21 @@ def createService(endpoint, role, endpoint_config):
...
-for endpoint_config in config.main.bouncer_endpoints:
+if config.main.tor_hidden_service and \
+        config.main.bouncer_endpoints is None and \
+        config.main.collector_endpoints is None:
+    bouncer_hsdir   = os.path.join(config.main.tor_datadir, 'bouncer')
+    collector_hsdir = os.path.join(config.main.tor_datadir, 'collector')
+    config.main.bouncer_endpoints   = [ {'type': 'onion', 'hsdir':   bouncer_hsdir} ]
+    config.main.collector_endpoints = [ {'type': 'onion', 'hsdir': collector_hsdir} ]
+
+for endpoint_config in config.main.get('bouncer_endpoints'):
```
hellais: You need to pass a default to `get()` (`get('bouncer_endpoints', [])`) otherwise you will attempt to iterate over a None when bouncer_endpoints is not defined.

hellais: I am going to fix this as part of merging it.

PR #78, opened 2016-05-23
Document configuration

joelanders:
https://github.com/TheTorProject/ooni-backend/issues/72

(Made this with Github's editor, which autogens the medium-helpful "Update README.rst" commit message.)

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -107,6 +107,23 @@ this would simply be ``'report'``) and ``tor_datadir`` (where you would
 like the spawned Tor process to keep its data). If you compiled Tor yourself,
 you'll likely want to specify it for the ``tor_binary`` option.
 
+Configure bouncer and collector endpoints
+.........................................
+
+The bouncer and collector are HTTP applications ("protocols" in twisted terminology) that can be configured to run on top of plain TCP, TLS, or onion service endpoints.
+Here is an example of the relevant part of the configuration::
+    bouncer_endpoints:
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -107,6 +107,23 @@ this would simply be ``'report'``) and ``tor_datadir`` (where you would
...
 
+Configure bouncer and collector endpoints
+.........................................
+
+The bouncer and collector are HTTP applications ("protocols" in twisted terminology) that can be configured to run on top of plain TCP, TLS, or onion service endpoints.
+Here is an example of the relevant part of the configuration::
+    bouncer_endpoints:
+    - {type: tls, port: 10443, cert: "private/ssl-key-and-cert.pem"}
+    - {type: tcp, port: 10080}
+    - {type: onion, hsdir: "/some/private/bouncer"}
```
bassosimone: Isn't the above YAML? If so, shouldn't the three lines above be further indented on the right?

joelanders: http://stackoverflow.com/questions/17014460/yaml-indentation-for-array-in-hash

I've seen it both ways. Do you prefer indented?

bassosimone: I didn't know one could avoid indenting: I pythonically assumed one had to :-).

> Do you prefer indented?

No, no need to make changes: I think whatever is compliant with YAML and consistent with the rest of the file is OK.

joelanders:
Added one more commit removing the old-style `tor_hidden_service` config option from the example config file.

hellais:
This looks good. Thanks!

PR #79, opened 2016-05-23
pass tor_binary config to txtorcon

joelanders:
Verified you can compile a tor2web tor, set `tor_binary` to point to it, set `tor2webmode` to true, and it works.


joelanders:
Fixed and checked.

PR #80, opened 2016-05-26
Fix privilege shedding and daemonisation of oonibackend

hellais:
* Rewrite to use twistd

* Launch tor only after we have forked

* Set the UID of when starting tor


hellais:
I am going to proceed in merging this so that I can begin making a new release of ooni-backend that includes this.

It's needed to deploy on mlab, for example.

data/archive/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
bassosimone: wow, didn't know this was possible!

bassosimone:
For the records, I did read this diff and it looks good!

PR #84, opened 2016-05-30
Vagrant: use box canonical name rather than URL

bassosimone:
This allows us to get updates running `vagrant box update`.

While at it, add `/.vagrant/` to `.gitignore`.


hellais:
Sounds good!

PR #85, opened 2016-05-30
Map also the connect_error

hellais:


oonib/testhelpers/http_helpers.py
```diff
@@ -390,6 +394,9 @@ def tcp_connect(self, socket):
         except ConnectionRefusedError:
             socket_info['status'] = False
             socket_info['failure'] = 'connection_refused_error'
+        except ConnectError:
+            socket_info['status'] = False
+            socket_info['failure'] = 'connect_error'
```
bassosimone: Reading what you wrote in TheTorProject/ooni-probe#457, I understand that you have seen the "no route to host" error. Perhaps, it would be interesting if such network error could become the error message as opposed to the more general "connect_error". Saying this because "no route to host" tells actually some precise information on how the host is censored, while "connect_error" strikes me as too broad. Is there a way (and/or a spec) to add information to the "failure" string? Meaning: I am fine with "connect_error" as the first cut error, but I'd rather also see the exact error afterwards.

hellais: Yes I agree that we should do a better job at extracting the error message, however mapping them to something that is actually parseable is a bit of work and it would need to be done in both the probe and in the backend.

The actual error string is set, I believe, by the underlying `socket` and not by twisted itself so that is where we would have to go hunting for the error strings.

bassosimone: Alright, let's open a ticket for this then. I think we should sit sometime in the future and think at how this could be done portably and robustly including also MK into the reasoning.

bassosimone: Ticketized as TheTorProject/ooni-spec#64 (which is a nice number for a ticket!)

PR #89, opened 2016-06-02
Add note about limitations of the gen-ssl-key-cert

hellais:


anadahz:
I think it makes sense to makes this clear in the README documentation as well and perhaps add documentation on how one can test this without ooni-probe.

hellais:
Can we merge this? I think it's better than nothing.

anadahz:
Sure, merging.

Issue #86, opened 2016-06-01
Support saving state when ooni-backend is stopped and restarted


hellais:
Currently if you stop an instance of ooni-backend and then restart it, the reports that are in progress will be lost.
That is if a client has created a report and has not closed it and the backend is taken down, when it goes back online they will no longer be able to update the report with that given ID.

The main issue has to do with the `reports` data structure that is used to handle the lifecycle of a report that is currently just in memory (https://github.com/TheTorProject/ooni-backend/blob/master/oonib/report/handlers.py#L213).

In order to implement a fix for this I see various possible options that go from easiest to implement, but less robust, to hardest but stronger:

* When the backend goes down pickle `reports` (https://github.com/TheTorProject/ooni-backend/blob/master/oonib/report/handlers.py#L200) to a certain file and when it comes back up load it from the pickle.

* Instead of checking the existence of a report with a certain ID based on the in-memory `reports` dictionary we should do it by checking if a file in the `raw_reports` directory exists.

* Move all the logic of creation and updating of reports into the database

hellais:
This is now implemented.

PR #91, opened 2016-06-08
Move state of reports to the filesystem

hellais:
This implements one of the solutions outlined in #86 to support saving the state of a ooni-backend instance when it is shutdown.

Basically we have two files for an opened report with ID `$REPORT_ID` they both reside in the `$REPORTS_DIR`.

1. The file at the path `$REPORTS_DIR/$REPORT_ID` is the actual partial report, that get's updated every time we need to create a new entry. 

2. The file at the path `$REPORTS_DIR/$REPORT_ID-metadata.json` contains the metadata associated with this report (the metadata sent on `open()`).

Every time we get a new update request we append to file 1. and touch file 2.

At regular intervals we look inside of the `$REPORTS_DIR` for all files ending with `-metadata.json` and look at their `mtime` with `stat`. If they have not been modified in a time frame greater than the stale_time we consider them stale and close them.

### DIFF
--- a/oonib/api.py
+++ b/oonib/api.py
@@ -1,29 +1,43 @@
 from cyclone import web
 
-from oonib.main.api import mainAPI
-from oonib.deck.api import deckAPI
-from oonib.report.api import reportAPI
-from oonib.input.api import inputAPI
-from oonib.policy.api import policyAPI
 from oonib.bouncer.api import bouncerAPI
 
 from oonib.config import config
 
-ooniBouncer = None
-ooniBackendAPI = []
-ooniBackendAPI += reportAPI
+class OONICollector(web.Application):
+    def __init__(self):
+        from oonib.main.api import mainAPI
+        from oonib.deck.api import deckAPI
+        from oonib.report.api import reportAPI
+        from oonib.input.api import inputAPI
+        from oonib.policy.api import policyAPI
+        from oonib.report.handlers import checkForStaleReports
 
-if config.main.input_dir:
-    ooniBackendAPI += inputAPI
+        handlers = []
+        handlers += reportAPI
 
-if config.main.deck_dir:
-    ooniBackendAPI += deckAPI
+        if config.main.input_dir:
+            handlers += inputAPI
 
-if config.main.policy_file:
-    ooniBackendAPI += policyAPI
+        if config.main.deck_dir:
+            handlers += deckAPI
 
-if config.main.bouncer_file:
-    ooniBouncer = web.Application(bouncerAPI, debug=True, name='bouncer')
+        if config.main.policy_file:
+            handlers += policyAPI
 
-ooniBackendAPI += mainAPI
-ooniBackend = web.Application(ooniBackendAPI, debug=True, name='collector')
+        handlers += mainAPI
+
+        checkForStaleReports()
+
+        web.Application.__init__(self, handlers, name='collector')
+
+
+class OONIBouncer(web.Application):
+    def __init__(self):
+        handlers = []
+        handlers += bouncerAPI
+
+        # Follows the same pattern as the above so we can put some
+        # initialisation logic for the bouncer as well in here perhaps in
+        # the future.
+        web.Application.__init__(self, handlers, name='bouncer')
--- a/oonib/api.py
+++ b/oonib/api.py
@@ -1,6 +1,5 @@
 from cyclone import web
 
-from oonib.bouncer.api import bouncerAPI
 
 from oonib.config import config
 
@@ -34,6 +33,8 @@ class OONICollector(web.Application):
 
 class OONIBouncer(web.Application):
     def __init__(self):
+        from oonib.bouncer.api import bouncerAPI
+
         handlers = []
         handlers += bouncerAPI
 
--- a/oonib/oonibackend.py
+++ b/oonib/oonibackend.py
@@ -11,7 +11,7 @@ import os
 import sys
 
 
-from oonib.api import ooniBackend, ooniBouncer
+from oonib.api import OONICollector, OONIBouncer
 from oonib.config import config
 from oonib.onion import get_global_tor
 from oonib.testhelpers import dns_helpers, ssl_helpers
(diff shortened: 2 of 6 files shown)


bassosimone:
Epic! Will review and test this!

oonib/api.py
```diff
@@ -1,29 +1,43 @@
 from cyclone import web
 
-from oonib.main.api import mainAPI
-from oonib.deck.api import deckAPI
-from oonib.report.api import reportAPI
-from oonib.input.api import inputAPI
-from oonib.policy.api import policyAPI
 from oonib.bouncer.api import bouncerAPI
```
bassosimone: For symmetry with moving APIs in class scope you would probably want to move also this statement

oonib/oonibackend.py
```diff
@@ -64,9 +64,9 @@ def getEndpoint(endpoint_config):
 
 def createService(endpoint, role, endpoint_config):
     if role == 'bouncer':
-        factory = ooniBouncer
+        factory = OONIBouncer()
```
bassosimone: Earlier `ooniBouncer` was only instantiated if `config.main.bouncer_file`, now it is instantiated here if `role == 'bouncer'`. This comment is just to check that this change has no unexpected implications.

oonib/report/handlers.py
```diff
@@ -192,27 +140,63 @@ def parseNewReportRequest(request):
 
     return parsed_request
 
-class ReportHandler(OONIBHandler):
+def closeReport(report_id):
+    report_dir = FilePath(config.main.report_dir)
+    archive_dir = FilePath(config.main.archive_dir)
+    report_path = report_dir.child(report_id)
+    report_metadata_path = report_dir.child(report_id +
+                                            "-metadata.json")
```
bassosimone: Maybe `"-metadata.json"` could be a global variable? (Would be a define in C)

oonib/report/handlers.py
```diff
@@ -192,27 +140,63 @@ def parseNewReportRequest(request):
...
+    print "Checking if %s exists and %s exists" % (report_path,
+                                                   report_metadata_path)
```
bassosimone: Could you use the logger here? Python3 will arrive sooner or later and that day we'll be happier if there are less instances of the print statement to forward port to Python3.

oonib/oonibackend.py
hellais: Good call, though I realised while refactoring this part that there was a redundant check for `config.main.bouncer_file` since it's existence is already tested down here: https://github.com/TheTorProject/ooni-backend/pull/91/files/89e2e4628198bf6f9cb13f0732c774ca0652324f#diff-47300978dd2c3f736ff16c63290517aaR157.

I thought a bit about what the possible consequences for only trapping it there are and I didn't see any case where it would be a problem. Though it's perhaps useful to still think of some possible edge case I may not have thought of.

oonib/report/handlers.py
```diff
@@ -192,27 +140,63 @@ def parseNewReportRequest(request):
...
+    if not report_path.exists() or \
+            not report_metadata_path.exists():
+        raise e.ReportNotFound
+
+    with report_metadata_path.open('r') as f:
+        report_details = json.load(f)
```
bassosimone: I see that the case where `json.load()` throws `ValueError` because the metadata file is corrupted is not handled. Do we want to handle this case? (I think it might happen if we close when writing a metadata file is in progress.)

bassosimone: Ah, probably it is not handled because the stale report will eventually be reaped?

oonib/report/handlers.py
```diff
@@ -192,27 +140,63 @@ def parseNewReportRequest(request):
...
+    report_dir = FilePath(config.main.report_dir)
+    for report_metadata_path in \
+            report_dir.globChildren("*-metadata.json"):
+        last_updated = time.time() - report_metadata_path.getModificationTime()
+        if last_updated > config.main.stale_time:
+            report_id = report_metadata_path.basename().replace(
+                '-metadata.json', '')
+            closeReport(report_id)
+    reactor.callLater(config.main.stale_time,
+                      checkForStaleReports)
```
bassosimone: I suggest to schedule the callback as the first action of the callback such that, even in case of unexpected exception in the following code, we are robust in that the next scheduled callback will be called

bassosimone:
- [ ] diff looks good
- [ ] testing using MK as input
- [ ] testing using old YAML format
- [ ] ^Z of OONI during deck, shutdown and restart of oonib

oonib/report/handlers.py
hellais: Yes this was actually some left-over from debugging and it should for sure be moved to either a log.debug() or removed.

oonib/report/handlers.py
hellais: This is a fair point, however if the metadata file is corrupt we actually never end up cleaning it.

I think that this sort of race condition is not very likely to occur given that in order for it to happen you must have created a report without ever writing data to it.

I am going to handle this condition by deleting the metadata file as well as the report file and logging this with level warning.

oonib/report/handlers.py
hellais: Yes this is a very good idea.

hellais:
I believe the issue lies in the fact that ooniprobe 1.3.2 doesn't support non httpo collectors (i.e. https or http). I would advise you run it by specifying the collector address in httpo format.

bassosimone:
> I would advise you run it by specifying the collector address in httpo format.

Yes, this strategy is working, thanks!

oonib/oonibackend.py
bassosimone: Reading through the code; will report back soon.

bassosimone: >> [...] it's perhaps useful to still think of some possible edge case I may not have thought of.

> Reading through the code; will report back soon.

I do not see obvious way in which the discussed change could break existing behavior.

bassosimone:
I've done more testing! Going to merge!

PR #93, opened 2016-06-21
Make it clearer how to create the cert for TLS support

hellais:
This is probably useful for doing: https://github.com/TheTorProject/ooni-sysadmin/issues/61


anadahz:
I believe that it makes more sense to add a script or documentation on how to create certificates that can be interpreted and used by ooniprobe i.e letsencrypt, otherwise it doesn't make much sense as it confuses any person that would like to test the TLS bouncer, collector, test-helps backend functionality

bassosimone:
(I think this should be closed.)

PR #95, opened 2016-06-22
Implement fix to serve also intermediate certificates

hellais:
* This also changes the configuration to have to more options for tls
 endpoints (privkey and fullchain).

@anadahz noticed that with openssl versions > 0.9.8 if the server does not serve the full certificate chain it will fail validation.

This fix add support for specifying the full chain.


bassosimone:
```C++
REQUIRE(to_english("aho', a occhio compila") == "Looks good to me");
```

PR #97, opened 2016-06-24
Feature/log format

hellais:
This implements what is described in https://github.com/TheTorProject/ooni-backend/issues/96

@anadahz


hellais:
I am going to proceed with the self merge of this.

PR #101, opened 2017-01-26
Strip invalid tcp_connect fields in the web_connectivity test helper

hellais:
This is the ooni-backend side fix of:
https://github.com/TheTorProject/ooni-probe/issues/727


oonib/tests/test_bouncer.py
```diff
@@ -72,6 +72,22 @@
     test-helper: {fake_test_helper: 'fake_hostname'}
 """
 
+production_bouncer_file = """
+collector:
+  httpo://ihiderha53f36lsd.onion:
+    collector-alternate:
+    - {address: 'https://a.collector.ooni.io:4441', type: https}
```
bassosimone: `https://b.collector.ooni.io`?

oonib/tests/test_bouncer.py
```diff
@@ -72,6 +72,22 @@
...
+    - {address: 'https://das0y2z2ribx3.cloudfront.net', front: a0.awsstatic.com, type: cloudfront}
+    test-helper: {dns: '213.138.109.232:57004', http-return-json-headers: 'http://38.107.216.10:80',
+      ssl: 'https://213.138.109.232', tcp-echo: 213.138.109.232, traceroute: 213.138.109.232,
+      web-connectivity: 'httpo://7jne2rpg5lsaqs6b.onion'}
+    test-helper-alternate:
+      web-connectivity:
+      - {address: 'https://a.web-connectivity.th.ooni.io:4442', type: https}
```
bassosimone: IIRC `https://b.web-connectivity.th.ooni.io`?

oonib/tests/test_bouncer.py
hellais: Actually the canonical bouncer is currently using this address. Though anyways what is in here doesn't really matter for the purpose of the unittest.

oonib/tests/test_bouncer.py
hellais: Same thing as above. The canonical bouncer is still advertising the old address. We should switch it though when we update to the version in here.

oonib/tests/test_bouncer.py
bassosimone: (I don't think it's relevant, however)

oonib/tests/test_bouncer.py
bassosimone: (ditto)

PR #103, opened 2017-01-29
Fix backward compatibility with older clients

hellais:
Basically older clients expect to find a key with the socket name in the `tcp_connect` dictionary even when it's not a valid socket (i.e. a hostname).

To support these older clients entirely we need to return a key that contains the value `None`. If we set it to `None` we are able to:

1) Distinguish these on the analysis side
2) Older clients will treat this as if the endpoint were not blocked in all cases

I believe this is the correct behaviour and should not add any errors in interpreting the results.


hellais:
I am going to proceed with merging this so that we can cut a new release and fix compatibility with older clients.

PR #104, opened 2017-01-31
Fix/redirect

hellais:


hellais:
I am going to proceed with merging this.

PR #111, opened 2017-09-26
fix(report/handlers): accept more semver versions

bassosimone:
The current problem we have, specifically, is that since v1.2.0 the
mobile apps emit full semver including also the build.

This requires several parts of our infrastructure to improve the
regexp being used so to validate also this kind of input.

See also: measurement-kit/measurement-kit#1388

An earlier version of this diff was blessed by @hellais on Slack and
since then I just changed comments.


bassosimone:
The build is failing like this:

```
function killitwithfire () {

    trap - ALRM

    kill -ALRM $prog 2>/dev/null

    kill -9 $! 2>/dev/null && exit 0

}

function waitforit () {

    trap "killitwithfire" ALRM

    sleep $1& wait

    kill -ALRM $$

}

waitforit $1& prog=$! ; shift ;

+prog=6047

+shift

+waitforit

trap "killitwithfire" ALRM INT

+trap killitwithfire ALRM

+trap killitwithfire ALRM INT

"$@"& wait $!

+wait

+sleep

+wait 6049

sleep: missing operand

Try 'sleep --help' for more information.

RET=$?

+RET=0

if [[ "$(ps -ef | awk -v pid=$prog '$2==pid{print}{}')" != "" ]]; then

    kill -ALRM $prog

    wait $prog

fi

+kill -ALRM 6046

ps -ef | awk -v pid=$prog '$2==pid{print}{}'

killitwithfire

++killitwithfire

++trap - ALRM

++kill -ALRM 6047

++kill -9 6049

++ps -ef

++awk -v pid=6047 '$2==pid{print}{}'

/home/travis/.travis/job_stages: line 57:  6046 Segmentation fault      (core dumped) ./.travis.test.sh 30 ./bin/oonib

The command "chmod +x .travis.test.sh && ./.travis.test.sh 30 ./bin/oonib" exited with 139.
```

I don't understand very well what it could be, and it seems like this is something related to the way in which travis is handling the test. I am going to restart it to see if it's deterministic (I actually hope so).

bassosimone:
I am experiencing more errors with the build, like that the keyserver doesn't know the key. 😡

bassosimone:
Alright, it seems it's time to play golf.

bassosimone:
I really _hate_ these coveralls annoying comments that don't serve any purpose.

bassosimone:
In the end, I decided to rewrite the test that caused a segfault.

I believe the problem appeared now that travis has upgraded its infrastructure to 14.04: the previous travis build was for 03d738b and occurred on Apr, 6 when travis was still using 12.04.

The original script was brilliant: it did two nested waits to make sure the running process was either killed gracefully or with fire. But probably was too brilliant and triggered some edge case.

I did not want to wrestle too much with travis. Also, reading the script, it seems to me it's fine to rewrite it such that, if we cannot kill the background process, [the build will hang and then fail (on travis)](https://travis-ci.org/TheTorProject/ooni-backend/builds/279910733).

I guess having the local build hang and the travis build fail is good enough for our purposes.

bassosimone:
Alright, I have read the diff once more. I am going to self bless this as good, given that @hellais already blessed the diff improving the regexp on Slack and that the test changes "look good to me".

theres-no-place-like-127 0 0 1_www fullhdwpp com_

PR #116, opened 2019-02-19
Android 2.0.0 block

hellais:


oonib/report/handlers.py
```diff
@@ -138,6 +138,12 @@ def validateHeader(header):
     except KeyError:
         pass
 
+    # This is to not accept new measurements coming from OONI Probe Android
+    # 2.0.0, which was affected by bug:
+    # https://github.com/ooni/probe-android/issues/188
+    if header.get('software_name') == 'ooniprobe-android' and header.get('software_version') == '2.0.0':
+        raise e.InvalidRequestField('software_version')
+
```
darkk: I'd suggest `'2.0.0' or startswith('2.0.0-')` as betas were also affected and we can't retract binaries from the internet.
That's would also mimic [the crutch in the pipline](https://github.com/ooni/sysadmin/commit/1c2dac0fd19e54a00b59a029a87715143fd40ec5#diff-5989d9055652f192d4224585e196d40dR25).

hellais: Good point. I will do that.

hellais:
I am going to proceed with merging this and work towards releasing a new ooni-backend version that includes this fix.

PR #118, opened 2019-03-14
bouncer: implement /collector and /test-helpers

bassosimone:
This allows us to deploy a bouncer with a much simpler API. So, we can
start using such better API in apps. This makes us ready to switch to
use the OONI orchestra rather than legacy services, since the API added
here is the same API of some OONI orchestra endpoints.

### DIFF
--- a/oonib/bouncer/handlers.py
+++ b/oonib/bouncer/handlers.py
@@ -190,6 +190,72 @@ class Bouncer(object):
             nettests.append(nettest)
         return {'net-tests': nettests}
 
+    def formatCollectorsWithoutPolicy(self):
+        ''' Formats the collectors without policy for the new
+            /api/v1/collectors endpoint. '''
+        results = []
+        for collector in self.knownCollectorsWithoutPolicy:
+            results.append({'type': 'onion', 'address': collector})
+            section = self.bouncerFile.get('collector')
+            if not section: continue
+            info = section.get(collector)
+            if not info: continue
+            for alt in info['collector-alternate']:
+                r = {}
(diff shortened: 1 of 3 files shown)


oonib/bouncer/handlers.py
```diff
@@ -190,6 +190,68 @@ def filterByNetTests(self, requested_nettests):
...
+            section = self.bouncerFile.get('collector')
+            if not section: continue
+            info = section.get(collector)
+            if not info: continue
+            for alt in info['collector-alternate']:
+                r = {}
+                for k in ('type', 'address', 'front'): r[k] = alt.get(k)
+                if not r['type'] or not r['address']: continue
+                if not r['front']: del r['front']
+                results.append(r)
```
hellais: The section appears to be duplicated in the two sections, perhaps it can be abstracted as a function called something like `get_alternate_address()`?

hellais:
I left a minor nitpick quality, feel free to address it or ignore it.

I appreciate the fact that there are unittests for this bit of code and that makes me comfortable with accepting this (even if the minor nitpick is not addressed) and deploying it.

oonib/bouncer/handlers.py
bassosimone: Done!

bassosimone:
> I left a minor nitpick quality, feel free to address it or ignore it.

I implemented your suggestion. Please, take another quick look. If all is good, I'll tag a new release and proceed with deploying it next week. Thanks.

bassosimone:
Thanks!

PR #123, opened 2019-06-10
/api/v1/{bouncer,test-helpers}: no duplicates

### LINKED ISSUE #120: bouncer: don't send duplicate information
Currently we don't deduplicate the information we send when bouncing; see:

```
$ curl -s https://bouncer.ooni.io/api/v1/collectors|python -m json.tool
[
    {
        "address": "httpo://42q7ug46dspcsvkw.onion",
        "type": "onion"
    },
    {
        "address": "https://c.collector.ooni.io:443",
        "type": "https"
    },
    {
        "address": "https://das0y2z2ribx3.cloudfront.net",
        "front": "a0.awsstatic.com",
        "type": "cloudfront"
    },
    {
        "address": "httpo://ganriuwqt5vgbgnj.onion",
        "type": "onion"
    },
    {
        "address": "https://b.collector.ooni.io:443",
        "type": "https"
    },
    {
        "address": "https://das0y2z2ribx3.cloudfront.net",
        "front": "a0.awsstatic.com",
        "type": "cloudfront"
    }
]
```

We should do that. Otherwise an app may be trying to connect to the same service more than once, which is a bit of a bummer.

bassosimone:
Closes #120

Lists are usually small (~3-5 elements), therefore I didn't felt
entitled to do any kind of smart filtering.

It would have been possible to fix duplications at a more fundamental
level, e.g., when loading. I did choose the path of not messing around
with existing code. Rather, I did just implement de-duplication when
we needed it. I am not 100% familiar with this codebase and didn't want
to perform more radical changes with possible side effects.

Because this product is approaching EOL and we should make sure we
make ooni/collector production-ready, I decided to just implement the
minimal set of changes to close the existing issue.

### DIFF
--- a/oonib/bouncer/handlers.py
+++ b/oonib/bouncer/handlers.py
@@ -204,7 +204,7 @@ class Bouncer(object):
                 r = self.format_alternate_address(alt)
                 if r:
                     results.append(r)
-        return results
+        return self.remove_duplicates_list(results)
 
     def formatTestHelpersWithoutPolicy(self):
         ''' Formats the test helpers without policy for the new
@@ -230,7 +230,7 @@ class Bouncer(object):
                         r = self.format_alternate_address(e)
                         if r:
                             results[k].append(r)
-        return results
+        return self.remove_duplicate_dict_values(results)
 
     @staticmethod
     def format_alternate_address(entry):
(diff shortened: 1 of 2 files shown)


bassosimone:
> [...] a ooni/bouncer microservice.

Code-wise, that would be my preference. Yet, reusing connections is also nice.

Thanks for the review.

PR #124, opened 2019-06-11
oonib: sketch out single measurement upload

bassosimone:


oonib/report/handlers.py
```diff
@@ -402,6 +408,37 @@ def get(self):
...
+                measurement.get('data_format_version', '0.2.0')
+            ),
+            'format': 'json',
+            'content': self.request.body,
+        })['report_id']
+        closeReport(reportID)
+        # TODO(bassosimone): the specification says that we should also
+        # return a measurement ID. Yet, it's not possible to add this
+        # functionality in here easily IMHO. So, should I also change the
+        # spec to say "SHOULD return a measurement ID"?
```
hellais: I don't think we should change that as it's pretty important that we start stamping the measurement_id inside of the backend ASAP.

I am of the opinion that we should roll out this single measurement API using the new ooni/collector code which already supports measurement_id stamping.

hellais: Actually we don't use the measurement_id as per the spec in https://github.com/ooni/pipeline/blob/master/docs/ooni-uuid.md#collector-stamped-ooid inside of the ooni/collector either.

At this point I am not sure what is the best way forward. I though think that once we start stamping the `measurement_id` in any collector the spec ought to say that it is strict requirement as this is a very important aspect of the collector which should not be optional.

bassosimone: "The legacy implementation is allowed to return an empty string here for backwards compatibility, however new implementations MUST return a valid OOID."

bassosimone: > I am of the opinion that we should roll out this single measurement API using the new ooni/collector code which already supports measurement_id stamping.

I disagree here. I'd like to deploy this new code in production when approved / next week, so the next iteration of mobile devices already takes advantage of this. I'd also like to deploy the new codebase as `d.collector.ooni.io` next week, without announcing it via bouncer, so we can start playing with it in our integration and automated tests.

hellais:
As discussed async I suggest doing the following to go forward WRT to this.

The assumption is that having a testing deployment of this feature is blocking on continuing work on the client:

1. A new docker image is created from this branch without merging it into master
2. Said image is deployed to a testing machine, perhaps `amstestcollector.ooni.nu`
3. Testing and development on the client can continue
4. This feature is implemented in ooni/collector (the code is there, the API endpoint just needs to be uncommented)
5. We deploy ooni/collector on a.collector and start trying it out with a subset of the clients and check if it works with the pipeline
6. We implement the measurement_id feature as per the spec
7. More testing
8. We roll this out in production and start using it for b.collector
9. More testing and assurance that nothing breaks
10. It's ready for prime time and backend-legacy/collector is deprecated 🎉

bassosimone:
> As discussed async I suggest doing the following [...]

Thanks for summarising our discussion. I am going to update the parent issue https://github.com/ooni/probe/issues/863 by merging the above into my original check-list.

This PR should stay open as draft, for documentation purposes, as long as we're using this code for testing. Once that testing is over, we should probably close it, as there seems to be no consensus on moving forward this prototype to become production code merged in master.

bassosimone:
Done here! https://github.com/ooni/probe/issues/863#issue-454133802

bassosimone:
I am going to close this PR, since it has not been merged in more than two years now and we are clearly not going to implement single measurement upload.

PR #646, opened 2023-03-19
Make the priority list public

hellais:
This is needed so we can allow listing priorities publicly

### DIFF
--- a/api/ooniapi/citizenlab.py
+++ b/api/ooniapi/citizenlab.py
@@ -794,7 +794,6 @@ def post_propose_changes() -> Response:
 
 
 @cz_blueprint.route("/api/_/url-priorities/list", methods=["GET"])
-@role_required(["admin"])
 def list_url_priorities() -> Response:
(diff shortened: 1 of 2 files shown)


api/ooniapi/citizenlab.py
```diff
@@ -794,7 +794,6 @@ def post_propose_changes() -> Response:
 
 
 @cz_blueprint.route("/api/_/url-priorities/list", methods=["GET"])
```
FedericoCeratto, line 796: Should we expose it in the public API instead?

hellais: I think we should keep it here since it's currently only used internally and I don't think we want to make any guarantees for future backward compatibility of this API.

hellais:
xref: https://github.com/ooni/test-lists-ui/pull/72

FedericoCeratto:
Rebased and updated chahngelog

FedericoCeratto (approved):

PR #652, opened 2023-03-22
Fix platform extraction

FedericoCeratto:


FedericoCeratto:
0.74~pr652-34 deployed on test

PR #653, opened 2023-03-23
Fix domain listing query

hellais:
A domain can appear in multiple country lists, so we need to pick the first occurrence of it sorted by country_code (so we give priority to the category code of the domain in the global list).


FedericoCeratto:
1.0.51~pr653-37 deployed on -test

FedericoCeratto (approved):

FedericoCeratto:
1.0.51~pr653-37 deployed on FSN

PR #654, opened 2023-03-31
Limit the count of measurements by test_name in domain list

hellais:
In the new domain pages we only display measurements coming from web_connectivity, yet in the listing we include in the count also non-web_connectivity measurements.
This PR is to address that so that we only limit it to those measurements.


FedericoCeratto (approved):

FedericoCeratto:
1.0.52~pr654-40 deployed on test

FedericoCeratto:
1.0.52~pr654-40 deployed on FSN

PR #692, opened 2023-06-30
Install certs for Vector in the TH rotation script

FedericoCeratto:


FedericoCeratto:
Deployed 1.10~pr692-102 on FSN

Issue #615, opened 2022-09-22
API: Filtering measurements by experiment_version  


arky:
It would be good to retrieve measurements using API by different  experiment versions.
This is really helpful during testing the experimental version of ooni-probe, miniooni where you can quickly find out the most common issues with such an API call.

PR #703, opened 2023-08-15
Filter list_measurement by versions

FedericoCeratto:
Adds filters for software_version, test_version, engine_version to list measurements
https://ams-pg-test.ooni.org/apidocs/#/default/get_api_v1_measurements
Related to https://github.com/ooni/backend/issues/615

### DIFF
--- a/api/ooniapi/measurements.py
+++ b/api/ooniapi/measurements.py
@@ -690,7 +690,6 @@ def list_measurements() -> Response:
     test_name = param("test_name")
     since = param_date("since")
     until = param_date("until")
-    since_index = param("since_index")  # unused
     order_by = param("order_by")
     order = param("order", "desc")
     offset = int(param("offset", 0))
--- a/api/ooniapi/measurements.py
+++ b/api/ooniapi/measurements.py
@@ -629,21 +629,21 @@ def list_measurements() -> Response:
         type: string
         description: |
           Set "true" for confirmed network anomalies (we found a blockpage, a middlebox, etc.).
-          Default: both true and false
+          Default: no filtering (show both true and false)
 
       - name: anomaly
         in: query
         type: string
         description: |
           Set "true" for measurements that require special attention (likely to be a case of blocking)
-          Default: both true and false
+          Default: no filtering (show both true and false)
 
       - name: failure
         in: query
         type: string
         description: |
           Set "true" for failed measurements (the control request failed, there was a bug, etc.).
-          Default: both true and false
+          Default: no filtering (show both true and false)
 
       - name: software_version
         in: query
--- a/api/ooniapi/measurements.py
+++ b/api/ooniapi/measurements.py
@@ -583,77 +583,66 @@ def list_measurements() -> Response:
       - name: report_id
         in: query
         type: string
-        description: The report_id to search measurements for
+        description: Report_id to search measurements for
       - name: input
         in: query
         type: string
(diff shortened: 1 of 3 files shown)


FedericoCeratto:
1.0.64~pr703-116 deployed on -test

api/ooniapi/measurements.py
```diff
@@ -800,6 +807,18 @@ def list_measurements() -> Response:
         query_params["test_name"] = test_name
         fpwhere.append(sql.text("test_name = :test_name"))
 
+    if software_version is not None:
+        query_params["software_version"] = software_version
+        fpwhere.append(sql.text("software_version IN :software_version"))
```
bassosimone, line 812: We're using `IN` rather than string equality because we want to match patterns--correct?

FedericoCeratto: `IN` matches the presence of a version in an array of accepted versions

FedericoCeratto: Done

api/ooniapi/measurements.py
```diff
@@ -583,77 +584,81 @@ def list_measurements() -> Response:
...
       - name: confirmed
         in: query
         type: string
-        collectionFormat: csv
-        items:
-          type: string
         description: |
-          Will be true for confirmed network anomalies (we found a blockpage, a middlebox was found, the IM app is blocked, etc.).
+          Set "true" for confirmed network anomalies (we found a blockpage, a middlebox, etc.).
+          Default: both true and false
```
bassosimone, line 632: How can the default be true and false at the same time?

api/tests/integ/test_integration.py
```diff
@@ -773,6 +773,34 @@ def test_list_measurements_ZZ(client):
     assert resp.status_code == 403
 
 
+def test_list_measurements_filter_software_version_nomatch(client):
+    # https://github.com/ooni/backend/issues/15
+    url = "measurements?since=2021-07-09&until=2021-07-10&software_version=9.9.9"
```
bassosimone, line 778: Because the functionality is advertised as a comma separated list, can all the tests here be using the comma separated version of the parameter rather than using just a scalar? 🙏

api/ooniapi/measurements.py
FedericoCeratto: When the value is not set it's handled as None and that generates an SQL query without filter.

bassosimone: Thanks for explaining! I still think the docs are a bit confusing. Maybe there is a better way to express this concept?

FedericoCeratto: Updated

api/tests/integ/test_integration.py
```diff
@@ -773,6 +773,55 @@ def test_list_measurements_ZZ(client):
...
+
+def test_list_measurements_filter_test_version(client):
+    # https://github.com/ooni/backend/issues/15
+    url = "measurements?since=2021-07-09&until=2021-07-10&test_version=0.4.0"
+    resp = api(client, url)
+    assert len(resp["results"]) == 100
+
+
+def test_list_measurements_filter_engine_version(client):
+    # https://github.com/ooni/backend/issues/15
```
bassosimone, line 819: Are you sure this is the correct issue URL?

bassosimone (approved):
LGTM, but I'd double check whether the issue URL you're using is correct, thanks!

🙏 🐳

Issue #608, opened 2022-08-11
pipeline: vanilla_tor scoring is broken


hellais:
Here is an example of a measurement of vanilla_tor being blocked: https://explorer.ooni.org/measurement/20220802T215209Z_vanillator_CN_17623_n1_LXFdyo7ao2kE7sQj, but the scores field taken from the API says otherwise: https://api.ooni.io/api/v1/measurement_meta?report_id=20220802T215209Z_vanillator_CN_17623_n1_LXFdyo7ao2kE7sQj.

```
{"anomaly":false,"category_code":null,"confirmed":false,"failure":false,"input":"","measurement_start_time":"2022-08-02T21:55:29Z","measurement_uid":"20220802215530.933188_CN_vanillator_f4e06257ed70aac2","probe_asn":17623,"probe_cc":"CN","report_id":"20220802T215209Z_vanillator_CN_17623_n1_LXFdyo7ao2kE7sQj","scores":"{\"blocking_general\":0.0,\"blocking_global\":0.0,\"blocking_country\":0.0,\"blocking_isp\":0.0,\"blocking_local\":0.0}","test_name":"vanilla_tor","test_start_time":"2022-08-02T21:52:07Z"}
```

FedericoCeratto:
Vanilla Tor scoring was added in https://github.com/ooni/pipeline/pull/242

FedericoCeratto:
Fixed

PR #706, opened 2023-08-18
Fix vanilla Tor scoring

FedericoCeratto:
The scoring logic is to be discussed. Once merged we can reprocess measurements and later on switch Explorer to use the scoring from the backend.
Related to https://github.com/ooni/backend/issues/608


fastpath/fastpath/core.py
```diff
@@ -914,27 +914,24 @@ def score_vanilla_tor(msm: dict) -> dict:
...
-    tor_log = tk.get("tor_log")
-    if tor_log is None:
+    tor_logs = tk.get("tor_logs") or tk.get("tor_log")
+    if tor_logs is None:
         # unknown bug
-        return scores
-
-    if (
-        "Bootstrapped 100%: Done" in tor_log
-        or "Bootstrapped 100% (done): Done" in tor_log
```
hellais, line 931: I believe this was in place to workaround a bug where older versions of probe did not have the `tor_log` key.

Also you need to handle both values of `tor_logs` and `tor_log`, since older versions of OONI Probe had a different name for that key, see: https://github.com/ooni/spec/blob/master/nettests/ts-016-vanilla-tor.md#incompatibility-with-ooniprobe-legacy

fastpath/fastpath/core.py
```diff
@@ -914,27 +914,24 @@ def score_vanilla_tor(msm: dict) -> dict:
     tk = g_or(msm, "test_keys", {})
     scores = init_scores()
 
-    nks = ("error", "success", "tor_log", "tor_progress_summary", "tor_progress_tag")
+    nks = ("error", "success", "tor_logs", "tor_progress_summary", "tor_progress_tag")
     if msm["software_name"] == "ooniprobe" and all_keys_none(tk, nks):
```
hellais, line 918: I think that since you are looking specifically for the `software_name` `ooniprobe` (i.e. the python implementation), you should be using the `tor_log` key instead of `tor_logs`, since that's the value used in legacy OONI Probe: https://github.com/ooni/spec/blob/master/nettests/ts-016-vanilla-tor.md#incompatibility-with-ooniprobe-legacy

hellais (changes requested):
Left some comments on bugs which should be fixed before merging.

fastpath/fastpath/core.py
FedericoCeratto: Updated!

fastpath/fastpath/core.py
FedericoCeratto: Updated as discussed.

hellais (changes requested):

fastpath/fastpath/core.py
```diff
@@ -914,7 +914,7 @@ def score_vanilla_tor(msm: dict) -> dict:
     tk = g_or(msm, "test_keys", {})
     scores = init_scores()
 
-    nks = ("error", "success", "tor_logs", "tor_progress_summary", "tor_progress_tag")
+    nks = ("error", "success", "tor_progress_summary", "tor_progress_tag")
```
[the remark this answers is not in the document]
FedericoCeratto: Check removed.

hellais (approved):
LGTM 🐳

PR #729, opened 2023-09-26
Remove duplicated line in debian/control

evilaliv3:
Just a simple correction removing a duplicated line in debian/control


FedericoCeratto:
Thanks!

Issue #709, opened 2023-08-24
The account registration endpoint should encode the email address in the link


hellais:
We would like to store the email in the incident reporting platform in the client so that we can populate a hidden field so that we can display the creator of an incident.

As such we would like to have the email link include an additional parameter with the email address of the user that created the token together with the `redirect_to` field.

The endpoint is `/api/v1/user_login`

hellais:
This has been implemented.

PR #732, opened 2023-09-27
Return email address on login

FedericoCeratto:
Implements https://github.com/ooni/backend/issues/709
Also add email address to incidents


FedericoCeratto:
1.0.73~pr732-160 deployed on -test

api/ooniapi/incidents.py
```diff
@@ -286,6 +306,8 @@ def post_update_incident(action: str) -> Response:
         if action == "create":
             incident_id = str(generate_random_intuid(current_app))
             req["id"] = incident_id
+            if get_client_role() != "admin" and mismatched_email_addr(req):
+                raise InvalidRequest
             if get_client_role() != "admin" and req["published"] == 1:
```
hellais, line 311: Shouldn't there be a check for `mismatched_email_addr` in here as well?

FedericoCeratto: In which line exactly? The check is being done on creation and update and not on deletion because, well, it's a deletion.

majakomel (approved):

PR #738, opened 2023-10-09
Test list HTML debug page

FedericoCeratto:
Implements https://github.com/ooni/backend/issues/739
The existing /api/_/debug_prioritization path returns JSON. This PR adds support for generating an HTML table that can be visualized in the browser to see what is in the test list.


FedericoCeratto:
Added related issue and description.

Issue #701, opened 2023-08-09
Add backend support for searching measurements from a specific OONI Run v2 link


hellais:
This would involve:
* [ ] Ensuring the OONI Run link ID is inside the measurement (cc @bassosimone )
* [x] Adding a new column in the DB to store it
* [x] Add support in fastpath
* [x] Add filtering in list_measurements

jbonisteel:
* Still missing right now - this work needs to be done - it is part of the grant 
* There is some work on the probe needed to unblock - @jbonisteel will coordinate with @bassosimone to create issue for this
* "filtering" by ID vs searching

jbonisteel:
TB discussed at backend meeting Oct 10

FedericoCeratto:
The PR was merged and deployed

bassosimone:
I documented the `ooni_run_link_id` annotation at https://github.com/ooni/spec/pull/280.

FedericoCeratto:
https://github.com/ooni/backend/pull/768 was merged and deployed

jbonisteel:
Closing this, we will re-open new tickets as needed if there are any issues

PR #768, opened 2023-11-03
Fastpath: Add ooni_run_link_id support

FedericoCeratto:
Implements: https://github.com/ooni/backend/issues/701
Specs: https://github.com/ooni/spec/pull/280/commits/220aaea0e76df21cb5f5218bb43d5d4a4219d306

### DIFF
--- a/fastpath/fastpath/core.py
+++ b/fastpath/fastpath/core.py
@@ -1662,7 +1662,17 @@ def process_measurement(msm_tup, buffer_writes=False) -> None:
         engine_name = g_or(annot, "engine_name", "")
         engine_version = g_or(annot, "engine_version", "")
         blocking_type = g(scores, "analysis", "blocking_type", default="")
+        ooni_run_link_id_str = g_or(annot, "ooni_run_link_id", "")
+        if ooni_run_link_id_str:
+            try:
+                ooni_run_link_id = int(ooni_run_link_id_str)
+            except ValueError:
+                ooni_run_link_id = None
+        else:
+            ooni_run_link_id = None
 
+        # TODO: build an object or typed dict here instead of passing args by
+        # position. Move the dict from db.py "new = dict(..."
         db.clickhouse_upsert_summary(
             measurement,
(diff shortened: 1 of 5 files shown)


bassosimone (approved):
🐳

hellais (approved):
lgtm

fastpath/fastpath/core.py
```diff
@@ -1662,7 +1662,17 @@ def process_measurement(msm_tup, buffer_writes=False) -> None:
...
         engine_version = g_or(annot, "engine_version", "")
         blocking_type = g(scores, "analysis", "blocking_type", default="")
+        ooni_run_link_id_str = g_or(annot, "ooni_run_link_id", "")
+        if ooni_run_link_id_str:
+            try:
+                ooni_run_link_id = int(ooni_run_link_id_str)
+            except ValueError:
+                ooni_run_link_id = None
+        else:
+            ooni_run_link_id = None
```
hellais, line 1672: Isn't it a bit more clear to rewrite this as:

```suggestion
        ooni_run_link_id_str = g_or(annot, "ooni_run_link_id", None)
        try:
            ooni_run_link_id = int(ooni_run_link_id_str)
        except (TypeError, ValueError):
            ooni_run_link_id = None
```

?

hellais (commented):
Made small suggestion to the diff. Otherwise LGTM

fastpath/fastpath/core.py
FedericoCeratto: It would be shorter but it would raise and catch an exception around int() even for the expected value `""`.

hellais: It would not raise the exception, since `int("")` throws a `ValueError` which is caught by the handler.

It's equivalent to the previous code.

FedericoCeratto:
0.84~pr768-221 deployed on -test

FedericoCeratto:
0.84~pr768-221 deployed on FSN

PR #791, opened 2024-01-18
2024 refactor

hellais:
This is a WIP PR. I mostly opened it so that I can test the CD/CI pipeline and so that I can have a branch I can pull into the devops tooling where I want to make use of it.

I will be updating it with more details once it ready for review.

Notes about the PR:
* Integrate CD/CI process for tier1 backend components
* Refactor tier1 components in separate service based on fastapi
* Make use of poetry for dependency management
* Implement Dockerfile for fastapi based backend

To preserve both functionality, for the moment the fastapi refactor duplicates a lot of the functionality of the existing API. Eventually all that code should be removed.


hellais:
Superseeded by #800 and #801

PR #797, opened 2024-02-01
Backend v2024.2.0

hellais:
This is a rebase of the https://github.com/ooni/backend/pull/791 branch following a discussion with @bassosimone to improve the legibility of the diff.


hellais:
We are closing this because in the end @bassosimone preferred to see the PRs structured differently.

PR #813, opened 2024-03-08
OONI Run v2 API improvements take 2

hellais:
This PR implements the changes discussed with @majakomel.

Highlights:
* `nettests` is now a reference to a separate table `oonirun_nettests` which contains the actual metadata of nettests
* `nettests` are validated much more strictly
* Tighter validation and typing is introduced as these fields have become database columns
* Versioning via the `revision` number is moved from the `oonirun` table into the `oonirun_nettest` table

An additional change that was introduced in this PR too is to rename `/api/v2/oonirun_links` endpoint into `/api/v2/oonirun-links`. For rational behind this, see the following resources:
* https://opensource.zalando.com/restful-api-guidelines/#129
* https://developers.google.com/search/docs/crawling-indexing/url-structure
* https://stackoverflow.com/questions/119312/urls-dash-vs-underscore
* https://stackoverflow.com/questions/10302179/hyphen-underscore-or-camelcase-as-word-delimiter-in-uris


hellais:
I did a force push to consolidate the git history into only 2 commits so that I can now rebase and merge.

ooniapi/services/oonirun/src/oonirun/routers/oonirun.py
```diff
@@ -0,0 +1,628 @@
...
+    )
+
+
+class OONIRunLink(OONIRunLinkBase):
+    oonirun_link_id: str
+    date_created: datetime = Field(
+        description="time when the ooni run link was created"
+    )
+    date_updated: datetime = Field(
+        description="time when the ooni run link was created"
```
majakomel, line 145: something like that?
```suggestion
        description="time when the ooni run nettest was last updated"
```

hellais: Good catch!

PR #825, opened 2024-03-15
Implement ooniauth API v2

hellais:
This adds a new endpoint which implements a slightly nicer ooniauth API.

The main changes are the following:
* POST `/v1/user_register` -> POST `/v2/ooniauth/user-login`
* GET `/v1/user_login` -> POST `/v2/ooniauth/user-session` with `login_token` in the request body
* GET `/v1/user_refresh_token` -> POST `/v2/ooniauth/user-session` with `authentication` header set
* GET `/_/account_metadata` -> GET `/v2/ooniauth/user-session`

So basically instead of 4 endpoints, we now have only 2:
* `/v2/ooniauth/user-login` (implements only POST)
* `/v2/ooniauth/user-session` (implements both POST and GET). The POST accepts a token in body the json body or the authorization header. Token in the request body takes priority over the authorization header.


hellais:
I will soon deploy this to dev and update this to dev and provide a link to the docs for the API.

majakomel (approved):

PR #828, opened 2024-03-21
Add support for setting and returning account_id

hellais:
@majakomel reported that we actually need the `account_id` for OONI Run to work.

We might eventually want to remove the redundancy by just storing email addresses directly.


DecFox (approved):
Changes look good to me! I am going to go ahead and merge

PR #829, opened 2024-03-21
fix(probe_services): disable torsf

bassosimone:
This diff uses richer input to disable the torsf experiment. This works under the assumption that probes periodically call check-in and refresh feature flags, which should be true for the probes maintained by the OONI team.

Mitigation for https://github.com/ooni/probe/issues/2692


hellais (approved):
LGTM

PR #830, opened 2024-03-21
wip: add experimental openvpn-config endpoint

ainghazal:
Add experimental entrypoint to provision credentials to OpenVPN experiment.

- Single provider (riseup) for the MVP.
- Append-only log for Clickhouse.
- Configurable "freshness" parameter to limit the query.
- Stub for inserting new credentials periodically.


hellais:
I added some patches that ports this over to the new pattern.

It's unfinished, but it should give you an idea of what's needed to get it to work.

Main things missing are:
1. Adding the alembic DB migrations (I changed the pattern slightly to work with multiple services)
2. Finishing the response model to support all that you need (I suppose you also need to do something to discover IPs and put them in the config payload, but I didn't see that in the original code either)
3. Write tests for everything

api/ooniapi/probe_services.py
```diff
@@ -661,10 +667,6 @@ def serve_tor_targets() -> Response:
     return nocachejson(torconf)
 
 
-# Unneded: we use an external test helper
-# @probe_services_blueprint.route("/api/private/v1/wcth")
-
-
```
hellais, line 667: You should remove from this patch anything that's touching the legacy API, since it's a bit scary to touch that code. We should keep the changes to that down to the bare minimum.

hellais:
Superseeded by: https://github.com/ooni/backend/pull/832

PR #831, opened 2024-03-22
fix: mark signal v0.2.4 as failed

bassosimone:
The https://github.com/ooni/probe-cli/pull/1421 PR trimmed the endpoints and bumped signal's version to v0.2.5.

So we need to ignore versions of signal lower than v0.2.5.

I am wondering whether we should also use a time window because otherwise what happens when we reprocess measurements. If my analysis is correct, then we have an additional issue that there's another check in this codepath that seems to mark signal measurements as failed without any temporal constraints. This would also cause reprocessing to cause downstream issues.

Part of https://github.com/ooni/probe/issues/2636

### DIFF
--- a/fastpath/fastpath/core.py
+++ b/fastpath/fastpath/core.py
@@ -1368,7 +1368,7 @@ def score_signal(msm: dict) -> dict:
             # https://github.com/ooni/probe/issues/2627
             scores["accuracy"] = 0.0
             return scores
-        
+
         if parse_version(tv) < parse_version("0.2.2"):
             start_time = msm.get("measurement_start_time")
             if start_time is None:
--- a/fastpath/fastpath/core.py
+++ b/fastpath/fastpath/core.py
@@ -1353,11 +1353,18 @@ def score_signal(msm: dict) -> dict:
     try:
         # https://github.com/ooni/probe/issues/2344
         tv = g_or(msm, "test_version", "0.0.0")
+
+        if parse_version(tv) <= parse_version("0.2.4"):
+            # See https://github.com/ooni/probe/issues/2636
+            # See https://github.com/ooni/probe-cli/pull/1421
+            scores["accuracy"] = 0.0
+            return scores
+
         if parse_version(tv) <= parse_version("0.2.3"):
             # https://github.com/ooni/probe/issues/2627
             scores["accuracy"] = 0.0
             return scores
-
+        
         if parse_version(tv) < parse_version("0.2.2"):
             start_time = msm.get("measurement_start_time")
             if start_time is None:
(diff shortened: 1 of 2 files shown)


fastpath/fastpath/core.py
```diff
@@ -1353,29 +1353,43 @@ def score_signal(msm: dict) -> dict:
...
+            and start_time >= datetime(2023, 11, 7)
+        ):
+            # See https://github.com/ooni/probe/issues/2636
+            # See https://github.com/ooni/probe-cli/pull/1421
+            scores["accuracy"] = 0.0
+            return scores
+
+        if parse_version(tv) <= parse_version("0.2.3") and start_time >= datetime(
+            2023, 11, 7
+        ):
```
bassosimone, line 1375: I see we're using the same `start_time` for both `if`s. I understand the `start_time` of 2023-11-07 comes from this issue https://github.com/ooni/probe/issues/2627. We noticed that we were testing too many endpoints on 2023-11-28, according to https://github.com/ooni/probe/issues/2636.

I think there are two options here:

1.

```suggestion
        if (
            parse_version(tv) <= parse_version("0.2.4")
            and start_time >= datetime(2023, 11, 7)
        ):
            # See https://github.com/ooni/probe/issues/2636
            # See https://github.com/ooni/probe-cli/pull/1421
```

2. Or using 2023-11-28 as the cutoff date for the fist check and dropping the `and parse_version(tv) > parse_version("0.2.3")` line.

The former suggestion coalesces two cases that are related together. The second one keeps them separate but applies the same kind of checking patterns to the code.

hellais: As discussed on slack it's probably better to be explicit and not collapse the two cases into one, so that it's more clear they are addressing specific different issues affecting different measurements.

fastpath/fastpath/core.py
```diff
@@ -1353,32 +1353,48 @@ def score_signal(msm: dict) -> dict:
...
+
+        msmt_start_time = msm.get("measurement_start_time")
+        if msmt_start_time is None:
+            scores["accuracy"] = 0.0
+            return scores
+        start_time = datetime.strptime(msmt_start_time, "%Y-%m-%d %H:%M:%S")
+
+        if parse_version(tv) <= parse_version("0.2.4") and parse_version(
+            tv
+        ) > parse_version("0.2.3"):
```
bassosimone, line 1365:
```suggestion
        if parse_version(tv) == parse_version("0.2.4"):
```

IIUC, we are saying that we're not sure about version 0.2.4 and we can just safely drop it, right? Then, in such a case, I suggest it's better to just check for equality here.

bassosimone (commented):
LGTM with an optional diff to further simplify the logic.

PR #832, opened 2024-03-28
Feature/riseupvpn config

hellais:
Create new ooniprobe service to support OpenVPN experiment.

Add experimental entrypoint to provision credentials to OpenVPN experiment.

* Single provider (riseup) for the MVP.
* Configurable "freshness" parameter to limit the query.
* Stub for inserting new credentials periodically.

This superseeds: https://github.com/ooni/backend/pull/830


ooniapi/services/ooniprobe/src/ooniprobe/utils.py
```diff
@@ -50,24 +51,26 @@ def fetch_openvpn_config() -> OpenVPNConfig:
...
-        )
-    ]
+    endpoints = []
+
+    r = httpx.get(RISEUP_ENDPOINT_URL)
+    r.raise_for_status()
+    j = r.json()
+    for ep in j["gateways"]:
+        ip = ep["ip_address"]
+        # TODO(art): do we want to store this metadata somewhere?
```
ainghazal, line 61: +1, I think the metadata as key-value can work for stuff like location, country code, etc coming from the provider... https://github.com/ooni/backend/pull/832/commits/c01c57636282a734b19da89bc777f370f5376b69#diff-a4debee7a43de30b4c89a21afca958ced896a3a313ebe595529ce0aead413cd8R48

for analysis it'll be useful to know when we first saw (start measuring) an endpoint, and when we ended serving it to clients.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v2.py
```diff
@@ -65,6 +71,7 @@ def update_vpn_provider(db: Session, provider_name: str) -> models.OONIProbeVPNP
             openvpn_key=vpn_cert["key"],
         )
         db.add(provider)
+        vpn_endpoints = fetch_openvpn_endpoints()
```
ainghazal, line 74: need try/except here too

ooniapi/services/ooniprobe/src/ooniprobe/routers/v2.py
```diff
@@ -114,6 +122,7 @@ def get_vpn_config(
             "cert": provider.openvpn_cert,
             "key": provider.openvpn_key,
         },
-        endpoints=[format_endpoint(provider.provider_name, ep) for ep in provider.endpoints],
+        # Pick 4 random endpoints to serve to the client
```
ainghazal, line 125: Picking 4 endpoints is great for MVP; perhaps we want to rethink the credential/endpoint turnover.
I think I would have `CREDENTIAL_FRESHNESS_INTERVAL_DAYS=1` so that we can exercise the mechanism more frequently.

How we conceptualize the choice is perhaps linked to the discussion in https://github.com/ooni/spec/pull/293#discussion_r1543574412 (i.e., a next step can be to select a bunch of IPs and measure all the port/transport combinations). The more I think about that discussion the more I'm inclined towards atomic ip-port-transport as the unit for a single measurement.

ainghazal (commented):
looks good to me, this is minimal and great to test the experiment; left some comments for future improvements since we have some of the discussion fresh.

the only change I see would be to reduce the frequency to a day (or less) so that we can test the updates.

to minimize migrations, perhaps it'd be good to deploy & test this branch and then pile up changes like metadata and perhaps a nicer abstraction for organizing measurement turnover. I'd suggest to have a conversation about that after landing the probe-cli experiment.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v2.py
hellais: It's actually correct here to not wrap it try, except, since this is the initial case when we don't have anything in the database, so failing to fetch the endpoints is a hard failure.

hellais: Maybe we should add a try, except block to the `        return update_vpn_provider(db, provider_name)
` call inside of `get_or_update_riseupvpn` where we could potentially return stale data in the event of an update failure.

PR #834, opened 2024-04-09
Fix/validation

hellais:
Pydantic has an unusual default behaviour when you set a `default` value for `Field`:

```
validate_default	bool | None	If True, apply validation to the default value every time you create an instance. Otherwise, for performance reasons, the default value of the field is trusted and not validated.	
```
https://docs.pydantic.dev/latest/api/fields/

This was leading to a validation bypass when the fields that specified an invalid default value were not being set.

We fix that bug, but also add extra validation that checks if the `author` matches the email address and refactoring of the authentication logic.


DecFox (approved):
LGTM!

PR #836, opened 2024-04-17
Remove rotation script and add 4.th to the roundrobin

hellais:
Changes:
* Remove the test helper rotation script
* Add 4.th.ooni.org to the round robin rotation rules


api/ooniapi/probe_services.py
```diff
@@ -515,8 +515,8 @@ def round_robin_web_test_helpers() -> List[Dict]:
         shift = q % 3 + 1
 
     out = []
-    for n in range(4):
-        n = (n + shift) % 4
+    for n in range(5):
+        n = (n + shift) % 5
```
bassosimone, line 519: This change is also making the new TH available. Is this intended? I thought the point was just to drop the TH rotation. Maybe consider updating the PR title as well.

hellais: Yes that's the addition of the 4.th.ooni.org to the rotation rules in the comment.

The idea is to start testing if 4.th.ooni.org is working (which is implemented in the new pattern) and then if all is good move to migrating 0-3 over to the new pattern iteratively.

I will update the PR title and description to make this more clear.

bassosimone (approved):
🐳

PR #838, opened 2024-04-23
Random sort of test helper addresses

hellais:
This PR adds support for randomly sorting the test helper addresses, which effectively leads to evenly distributing the load amongst them.


api/tests/unit/test_probe_services.py
```diff
@@ -0,0 +1,21 @@
+from unittest.mock import MagicMock, Mock
+
+import ooniapi.probe_services
+from ooniapi.probe_services import random_web_test_helpers, round_robin_web_test_helpers
+
+
+def test_web_test_helpers():
+    ooniapi.probe_services.extract_probe_ipaddr = Mock(return_value="1.2.3.4")
```
bassosimone, line 8: Does this change mean that other tests will have the mock configured or is there something I am missing that causes the mock to be reset once we exit this test?

hellais: fixed

bassosimone (approved):
🐳

PR #842, opened 2024-04-24
Correct the fixtures for scoring of signal bug probe#2627

hellais:
* The integration test was using as a fixture measurement from
  before the incident
(https://explorer.ooni.org/m/20210427000432.751743_AU_signal_490428d8ebc76e6b),
but was expecting it to be marked as failed
* The unit test was using a measurement that was from a newer version of
  probe
(https://explorer.ooni.org/measurement/20221118T104419Z_signal_IT_30722_n1_Q02UUAiiHlVU0VE6),
which should not be marked as failed either

As part of https://github.com/ooni/backend/pull/831 we improved the
logic for scoring to be more precise and so these checks were now
failing.

We now use in both cases a correct "bad case" and in the unit test we
also include the previous case as a "good case".


bassosimone (approved):
🐳

PR #845, opened 2024-04-26
feat: oonifindings service

### LINKED ISSUE #814: Port findings platform over to new pattern
This is about taking https://github.com/ooni/backend/blob/master/api/ooniapi/incidents.py and porting it over to make use of fastapi.

Here are the specs for how it should be:
* Service should be named `oonifindings`
* Service should be 100% identical to old `incidents`
* Service should follow the template of `oonirun`, but without postgresql and instead clickhouse
* Service should make use of the `incidents` tests to ensure it's working as intended. We should add more tests to reach 100% code coverage.
* Service should include metrics using prometheus
* We should also add an integration test that performs actions on the `incidents` backend-fsn API and compares the results with the new one (they should behave identically).
* Should setup `clickhouse` as a fixture in pytest by assuming the clickhouse server binary is installed on the system and available in `PATH`. It should start the server in a way that's it's throw away for the testing session, see how pytest-postgresql does this: https://github.com/ClearcodeHQ/pytest-postgresql

Once we have an initial iteration that's 100% compatible to the previous one we can work on a v2 version of the API that looks nicer, follows our more up to date API patterns and is backed by postgresql.

Keep an eye on this issue for docs on best-practices to come: https://github.com/ooni/backend/issues/817

DecFox:
Create oonifindings service with clickhouse support
Closes #814


ooniapi/services/oonifindings/tests/conftest.py
```diff
@@ -0,0 +1,100 @@
...
+    queries = sql_no_comment.split(";")
+    for q in queries:
+        q = q.strip()
+        if not q:
+            continue
+        click.execute(q)
+
+
+@pytest.fixture
+def clickhouse_migration(clickhouse):
```
hellais, line 28: We should for the time being swap this out for pytest-docker based approach:  https://github.com/ooni/data/blob/main/oonipipeline/tests/conftest.py#L39

ooniapi/services/oonifindings/src/oonifindings/routers/oonifindings.py
```diff
@@ -0,0 +1,459 @@
...
+        incident["published"] = bool(incident["published"])  
+        if account_id is None or client_role != "admin":
+            incident["email_address"] = ""
+    
+    setnocacheresponse(response)
+    incident_models = []
+    # TODO(decfox): try using OONIFindings.validate_model to populate model
+    for incident in incidents:
+        incident_model = OONIFinding(
+            incident_id=incident.id,
```
hellais, line 164: This should be changed to be `id` instead of `incident_id`

ooniapi/services/oonifindings/src/oonifindings/routers/oonifindings.py
```diff
@@ -0,0 +1,459 @@
...
+
+router = APIRouter()
+
+
+def utcnow_seconds():
+    return datetime.now(timezone.utc).replace(microsecond=0)
+
+
+class OONIFindingBase(BaseModel):
+    incident_id: str
```
hellais, line 37: You can use alias to change the serialization: https://docs.pydantic.dev/latest/concepts/fields/#field-aliases

ooniapi/services/oonifindings/src/oonifindings/routers/oonifindings.py
```diff
@@ -0,0 +1,459 @@
...
+    AND id = %(incident_id)s
+    AND creator_account_id != %(account_id)s
+    """
+    account_id = get_account_id_or_raise(authorization, jwt_encryption_key=jwt_encryption_key)
+    query_params = dict(incident_id=incident_id, account_id=account_id)
+    q = query_click(db, query, query_params)
+    return q[0]["cnt"] > 0
+
+
+class OONIFindingsUpdate(BaseModel):
```
hellais, line 340: Probably rename this to not conflict with the other `OONIFindingsUpdate`

DecFox:
I'll be rebasing this and pushing a new diff since we want to separate the pytest plugin work from the service

PR #849, opened 2024-05-08
feat: oonifindings service

DecFox:
This diff introduces the oonifindings revamp work with migration to fastapi and clickhouse support.


DecFox:
We can close this since this is from a fork not authorized to run tests on GH actions

Issue #855, opened 2024-06-21
Port findings platform to use postgresql


DecFox:
We want the findings platform to use postgresql to store ooni findings/incidents, since our clickhouse database is not backed up. This is part of https://github.com/ooni/backend/issues/814. We also want to replace the current API routes we have from `api/v1/incidents` to `api/v1/findings` as suggested here: https://github.com/ooni/backend/pull/850#discussion_r1634922142

PR #862, opened 2024-07-17
feat(oonifindings): add postgresql support

DecFox:
This diff replaces the clickhouse database used in the oonifindings (earlier incidents) service with a postgresql database. Closes https://github.com/ooni/backend/issues/855.


ooniapi/common/src/common/alembic/versions/a037e908f3a0_init_oonifindings.py
```diff
@@ -0,0 +1,53 @@
...
+        sa.Column("title", sa.String(), nullable=False),
+        sa.Column("short_description", sa.String(), nullable=False),
+        sa.Column("text", sa.String(), nullable=False),
+        sa.Column("reported_by", sa.String(), nullable=False),
+        sa.Column("email_address", sa.String(), nullable=False),
+        sa.Column("event_type", sa.String(), nullable=False),
+        sa.Column("published", sa.Integer(), nullable=False),
+        sa.Column("deleted", sa.Integer(), nullable=False, default=0),
+        sa.Column("CCs", sa.JSON(), nullable=True),
+        sa.Column("ASNs", sa.JSON(), nullable=True),
```
hellais, line 44: <bikeshed>
Can we use consistent capitalization for these columns?

I would suggest `country_codes` and `asns`
</bikeshed>

ooniapi/common/src/common/alembic/versions/a037e908f3a0_init_oonifindings.py
```diff
@@ -0,0 +1,53 @@
...
+
+def upgrade() -> None:
+    op.create_table(
+        "oonifinding",
+        sa.Column(
+            "incident_id",
+            sa.String(),
+            nullable=False,
+            primary_key=True,
+        ),
```
hellais, line 29: Can we add a column which is called something like finding_slug?

Also since we are now calling them findings, maybe this should be called `finding_id` or maybe even just `id`?

ooniapi/services/oonifindings/src/oonifindings/routers/v1.py
```diff
@@ -3,26 +3,26 @@
 """
 
 from datetime import datetime, timezone
-from typing import List, Dict, Optional, Union, Tuple, Any, Annotated
```
hellais, line 6: Should we maybe bump the version number on the router endpoint? As in make into `v2`. That way if we break something badly we can always revert back to speaking to the old backend as `v1`

ooniapi/services/oonifindings/src/oonifindings/routers/v1.py
```diff
@@ -122,7 +134,7 @@ class OONIFindingIncidents(BaseModel):
 @router.get(
     "/v1/incidents/search",
```
hellais, line 135: I think there is an opportunity here to improve a bit the semantics of these endpoints. We should speak to @majakomel about it though.

For example if we want to make this a truly restful endpoint, then we should not have the verb in the endpoint, but rather just have it be `/v2/findings?[query_params]`

ooniapi/services/oonifindings/src/oonifindings/routers/v1.py
```diff
@@ -131,55 +143,60 @@ def list_oonifindings(
...
-    incident_models = []
-    for i in range(len(incidents)):
-        incident = incidents[i]
-        incident_model = OONIFinding.model_validate(incident)
-        incident_models.append(incident_model)
-    return OONIFindingIncidents(incidents=incident_models)
+    return OONIFindingIncidents(incidents=findings)
 
 
 @router.get(
```
hellais, line 202: Same here, there is no need in RESTful API design to include the verb in the path, it can just be `/v2/finding/{finding_id | slug}`

ooniapi/services/oonifindings/src/oonifindings/routers/v1.py
```diff
@@ -191,212 +208,203 @@ def get_oonifinding_by_id(
...
+    if incident.end_time is not None:
+        incident.end_time = incident.end_time.replace(microsecond=0)
+        delta = incident.end_time - incident.start_time
+        if delta.total_seconds() < 0:
+            raise HTTPException(status_code=400, detail="invalid query paramters") 
+    return True
+
+
 @router.post(
     "/v1/incidents/create",
```
hellais, line 291: Again, no verb is needed, we can just call it POST `/v2/finding`

ooniapi/services/oonifindings/src/oonifindings/routers/v1.py
```diff
@@ -191,212 +208,203 @@ def get_oonifinding_by_id(
...
+    db.add(db_oonifinding)
+    db.commit()
+
     setnocacheresponse(response)
-    return OONIFindingsUpdateResponse(r=r, id=incident_id)
+    return OONIFindingsUpdateResponse(r=1, id=incident_id)
 
 
 @router.post(
     "/v1/incidents/update",
```
hellais, line 352: `PUT /v2/finding/`

ooniapi/services/oonifindings/src/oonifindings/routers/v1.py
```diff
@@ -191,212 +208,203 @@ def get_oonifinding_by_id(
...
 
-    update_request.creator_account_id = get_account_id_or_raise(
-        authorization, jwt_encryption_key=settings.jwt_encryption_key
-    ) 
-    incident_dict = prepare_incident_dict(update_request)
+    q = db.query(models.OONIFinding).filter(
+        models.OONIFinding.incident_id == incident_id
+    )
+    if token["role"] == "user":
+        q = q.filter(models.OONIFinding.creator_account_id == account_id)
```
hellais, line 372: I wonder if you can just do all of this in a single query, where you get one row and then apply the authentication checks once you have the row depending on the token role.

ooniapi/services/oonifindings/src/oonifindings/routers/v1.py
```diff
@@ -406,37 +414,38 @@ def update_oonifinding(
...
+    q.delete()
+    db.commit()
+
     setnocacheresponse(response)
     return {}
 
 
-
 @router.post(
     "/v1/incidents/{action}",
```
hellais, line 450: There is no need for this to be a separate endpoint, it can just be part of the update endpoint where an update request is made with the update column being set.

hellais (commented):
Note my comments are mostly related on how I think we ought to make the future of ooni findings look like.

I think we can start off by just swapping out the DB (as you did) and then improve the API based on my comments in a second moment, once we ensure that the existing API is 100% compatible with the existing one.

When and if we decide to do this, it's essential we also coordinate with @majakomel so that we don't break anything on the frontend.

ooniapi/common/src/common/alembic/versions/a037e908f3a0_init_oonifindings.py
DecFox: Done here: https://github.com/ooni/backend/pull/862/commits/af0efdee40cde5eda9eeca6504f72e2e913f55b6

ooniapi/common/src/common/alembic/versions/a037e908f3a0_init_oonifindings.py
DecFox: could you elaborate a little more on what the `finding_slug` column represents? made the `finding_id` change here: https://github.com/ooni/backend/pull/862/commits/2289ced2122ed6943c19bf58860806bb22bf0c97

ooniapi/services/oonifindings/src/oonifindings/routers/v1.py
DecFox: Thanks for the suggestion. You are right. I made some changes here: https://github.com/ooni/backend/pull/862/commits/f4c97afffc0abe206abbb457a6afddfe61684f78. In most cases, I have tried to assemble the query logic into a single query.

ooniapi/common/src/common/alembic/versions/a037e908f3a0_init_oonifindings.py
hellais: Basically we would like to have the ability to have some kind of `slug` for each finding so that instead of it being this opaque number for example `https://explorer.ooni.org/findings/45013413801` it should be possible to configure it be `https://explorer.ooni.org/findings/2024-07-bangladesh-blocks-facebook`. This needs to be stored in the DB and we need to have the ability to lookup both by number (we can assume a number is the old format) and then to a permanent redirect to the new slug based format (which is nicer for humans and search engines).

DecFox: thanks for explaining this! I added a nullable `finding_slug` column for now: https://github.com/ooni/backend/pull/862/files#diff-ebc4627fea3725c4e6f1dafa16bfb842ed5cfdb74c8b292e203fc07cd2581ecbR30 along with a `TODO` note in the models. We should be able to add the endpoint in the next version and also make this non-nullable (if required)

ooniapi/services/oonirun/tests/test_database.py
```diff
@@ -176,3 +176,37 @@ def test_upgrade_to_head(postgresql):
...
+def upgrade() -> None:
+    op.create_table(
+        "oonifinding",
+        sa.Column(
+            "incident_id",
+            sa.String(),
+            nullable=False,
+            primary_key=True,
+        ),
+        sa.Column("create_time", sa.DateTime(timezone=True), nullable=False),
```
hellais, line 189: I think you are missing the `finding_slug` here and you should rename the `incident_id` to `finding_id`, also the capitalization of the `CCs` `ASNs`, etc.

DecFox: thank you for pointing this out! This was an unrelated change we had to undo since this is the oonirun service. I have removed this from the diff now.

hellais (approved):
LGTM.

We can do all the v2 stuff later.

PR #869, opened 2024-08-10
fix: bypass health endpoint for initial deployment

DecFox:
The initial deployment cannot pass since we set environment variables after the deployment which are tested as part of `docker-smoketest.sh`. So we should comment the endpoint for the first deployment.


DecFox:
I'll merge this and we can then uncomment. The tests fail due to the commented out `/health`checkpoint.

Issue #911, opened 2025-01-09
Add more tests to the oonidata related endpoints in oonimeasurements service


hellais:
The oonidata components could benefit from having better testing: https://github.com/ooni/backend/tree/master/ooniapi/services/oonimeasurements/src/oonimeasurements/routers/data.

This would also give us more confidence in the deployment of the service working as intended: https://github.com/ooni/devops/issues/139.

As part of this it would probably be helpful to think about how to go about generating or provisioning test fixtures. One idea could be for the integration tests to make use of the actual oonipipeline to populate the tables inside of the oonidata service.

PR #921, opened 2025-01-15
Add more test to oonidata related endpoints

hynnot:
This PR is about the issue https://github.com/ooni/backend/issues/911

Steps:
- Add ARGS parameter in the make test command
- Add E2E tests to list analysis endpoint
- Add E2E tests to list observations endpoint
- Add E2E tests to aggregate analysis endpoint
- Add E2E tests to aggregate observations endpoint
- Refactor to add two fixtures in conftest to set the since and until params.
- Create a decorator to pass to int the str probe_asn filter.

Notes:
- An issue is created for the future so as not to depend on sample data and the necessary data is provided in each test to test the expected behavior: https://github.com/ooni/backend/issues/924


ooniapi/services/oonimeasurements/src/oonimeasurements/routers/data/utils.py
```diff
@@ -50,3 +52,11 @@ def test_name_to_group(tn):
         return "websites"
     # TODO(arturo): currently we only support websites
     return ""
+
+
+def parse_probe_asn(probe_asn):
```
hellais, line 57: Maybe it's a good idea to add typing to this argument to clarify that this function is correct only in the case that it's taking as argument `Union[str, int, None]` which is validated by the fastapi type annotation spec.

ooniapi/services/oonimeasurements/src/oonimeasurements/routers/data/utils.py
```diff
@@ -50,3 +52,11 @@ def test_name_to_group(tn):
...
+    if probe_asn.startswith("AS"):
```
hellais, line 58: I don't think this is correct. If `probe_asn` is None or `int`, which is supported by the spec, it will raise an exception in here.

ooniapi/services/oonimeasurements/tests/test_oonidata_aggregate_analysis.py
```diff
@@ -0,0 +1,130 @@
...
+    json = response.json()
+    assert isinstance(json["results"], list), json
+    assert len(json["results"]) > 0
+
+    for result in json["results"]:
+        assert "anomaly_count" in result, result
+        assert "domain" in result, result
+
+
+@pytest.mark.parametrize(
```
hellais, line 26: This is neat, I want' aware of it!

ooniapi/services/oonimeasurements/src/oonimeasurements/routers/data/utils.py
hynnot: Actually this is called after checking that it is not None always. About the int, I was testing and always I detected that really is a int as a string, this is the reason about the next lines. But maybe I don't find it the flow with int. Could you show this flow?

ooniapi/services/oonimeasurements/src/oonimeasurements/routers/data/utils.py
hynnot: Yes, but this is related to the following comment, at the very least it should be typed as a string. I will modify.

hellais: Can't it be an `int` though? In the previous version of this we were checking if it's `str` AND `startsWith(AS)` and then doing the casting to `int`: https://github.com/ooni/backend/pull/921/files/9d0c56f0f649a4b48bfa7b330d8f5c02a3b4d265#diff-e06ff97cf1030d17023c1d12d21537165829bb6cf53336da5cec4f95d43af526L100.

The fact it can be `Union[str, int, None]` comes from the types specified in the fastapi spec: https://github.com/ooni/backend/pull/921/files/9d0c56f0f649a4b48bfa7b330d8f5c02a3b4d265#diff-facb5f0fb26ad641ef437fcad578b9c36afe385edcf5cb27a3f590449c2ef856R171

hynnot: Yes, I think you are right! It should work with int too. But I think I found a bug with list_analysis regarding this. I'm going to create tests filtering with integers to verify that they should work correctly. I'll let you know when there are any changes.

hellais:
LGTM!

Issue #907, opened 2025-01-06
Port ams-ps probe-services related endpoints over to new pattern for deployment in ECS


hellais:
The `ams-ps.ooni.nu` host runs some the old OONI Orchestra component: https://github.com/ooni/orchestra.

In reality only the "registry" component of orchestra is being used, this is what the nginx configuration on backend-fsn looks like:
```
  location ~^/api/v1/(login|register|update) {
      proxy_http_version 1.1;
      proxy_set_header Host $http_host;
      proxy_set_header X-Real-IP $external_remote_addr;
      proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
      proxy_set_header X-Forwarded-Proto $scheme;
      proxy_read_timeout 900;

      proxy_pass https://registry.ooni.io:443;
  }
```

The goal of the registry is to provide a method for probes to register, login and update their probe metadata, such that we are able to implement access control for gaining access to the tor test helper configuration. This was implemented at a time when:
1. We needed the push notification tokens as we were using our own custom push notification solution instead of countly
2. We would be giving out private bridges to probes

Both of these are no longer true and so the login, register and update no longer serve a true purpose.

That said, probes expect these endpoints to work and so we still need to somehow serve responses for them otherwise they risk breaking. That's why we implemented in the API what is effectively a noop for these endpoints, however we haven't yet made the switch over to them (see: https://github.com/ooni/backend/blob/master/api/ooniapi/probe_services.py#L438).

Since we have quite a bit of work related to authenticating probes as part of the anonymous credentials project, it would probably be wise to decouple this piece from the monolith backend and integrate it into a separate component that we deploy using the [new pattern](https://docs.ooni.org/backend/ooniapi/).

This will set us up nicely to be able to then place the anonymous credentials code inside of it. Moreover, since probes are submitting metadata about their probe as part of this request, we can also use this endpoint to measure inconsistencies related to probe metadata and public IPs of probes (as per: https://github.com/ooni/ooni.org/issues/1646)

This is related to being able to discontinue ams-ps: https://github.com/ooni/backend/issues/892.

hellais:
We should put this inside of the OONI Run component: https://github.com/ooni/backend/tree/master/ooniapi/services/oonirun

PR #923, opened 2025-01-17
Port register, login and update services to the new ECS style

LDiazN:
This PR is related to #907 and ports services for login, register and update to the new ECS development style, mostly copied and pasted from the monolith, including tests. 

No host points to this services right now, so it shouldn't crash any existent probes or infrastructure. 

You can try the new endpoints [here](https://ooniprobe.dev.ooni.io/docs)

### DIFF
--- a/ooniapi/services/ooniauth/src/ooniauth/routers/v1.py
+++ b/ooniapi/services/ooniauth/src/ooniauth/routers/v1.py
@@ -115,6 +115,16 @@ async def user_login(
     settings: Settings = Depends(get_settings),
 ):
     """Auth Services: login using a registration/login link"""
+
+    # **IMPORTANT** You have to compute this token using a different key
+    # to the one used in ooniprobe service, because you could allow
+    # a login bypass attack if you don't. 
+    #
+    # The token used in ooniprobe is generated regardless of any authentication,
+    # because it's a toy token to please old probes. 
+    #
+    # We set this up in terraform
+
     try:
         dec = decode_jwt(
             token=token, key=settings.jwt_encryption_key, audience="register"


ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -0,0 +1,126 @@
...
+    token : str
+    expire : str
+
+@router.post("/login", tags=["ooniprobe"], response_model=ProbeLoginResponse)
+def probe_login_post(
+    probe_login : ProbeLogin, 
+    response : Response,
+    settings : Settings = Depends(get_settings),
+) -> ProbeLoginResponse:
+    global log
```
hellais, line 35: Is it necessary to make the `log` variable global?

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -0,0 +1,126 @@
...
+    payload = {"iat": issued_at, "aud": "probe_login"}
+    client_id = create_jwt(payload, key=settings.jwt_encryption_key)
+    log.info("register successful")
+
+    register_response = ProbeRegisterResponse(client_id=client_id)
+    setnocacheresponse(response)
+
+    return register_response
+
+class ProbeUpdate(BaseModel):
```
hellais, line 117: The ProbeUpdate request contains some useful metadata that we could use for the purpose of the anonymous credential anomaly detection to establish a baseline for comparing the probe_cc and probe_asn inconsistencies (see: https://github.com/ooni/orchestra/blob/master/registry/registry/handler/registry.go#L25).

Basically what I'm thinking of is that, if we assume that the update requests are not going to be tampered with as much as the measurement submit ones, we can use this to determine which inconsistencies are due to our geoip being off compared to the probe one, vs those which are due to the measurement metadata being inconsistent due to malice.

We can do parsing and logging of this as a follow up PR though.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -0,0 +1,126 @@
...
+    Note that most of the request body arguments are not actually 
+    used but are kept here to use the same API as the old version
+    """
+    global log
+
+    # client_id is a JWT token with "issued at" claim and
+    # "audience" claim. The "issued at" claim is rounded up.
+    issued_at = int(time.time())
+    payload = {"iat": issued_at, "aud": "probe_login"}
+    client_id = create_jwt(payload, key=settings.jwt_encryption_key)
```
hellais, line 109: We should take note of the fact the jwt_encryption_key is being shared for between probes and user authenticating via email in the auth service.

I don't think there is necessary a problem, given that the `aud` field is different, but we should make sure we don't forget to always check `aud` filed to avoid authentication bypasses.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -0,0 +1,126 @@
...
+from ...common.utils import setnocacheresponse
+
+router = APIRouter(prefix="/v1")
+
+log = logging.getLogger(__name__)
+
+class ProbeLogin(BaseModel):
+    # Allow None username and password 
+    # to deliver informational 401 error when they're missing
+    username : str | None = None
```
hellais, line 21: nitpick: Can you lint and reformat the code using `black` so that the spacing of the type hints is consistent with the rest of the codebase?

hellais:
This looks good.

Here is a summary of what should be addressed prior to merge:
* Few nitpick comments (global log and linting)
* Add comment in the auth component about the jwt token reuse
* Create follow up issue about recording metadata during update and adding support for metrics collection using prometheus or other metric collector

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
LDiazN: Not actually, I will remove it. I was mixing it with the pattern used in the monolith ([example here](https://github.com/ooni/backend/blob/9fa38c97c9f614e94a830234955c1aa0fe9157bc/api/ooniapi/citizenlab.py#L572)), but now that I look closer is very different

hellais (approved):

Issue #795, opened 2024-01-29
api list_measurements: add support for passing a list to filters 


hellais:
This is the list_measurements issue that implements what is outlined in here: https://github.com/ooni/backend/issues/584.

We should add support for passing in lists to every filter including the probe_cc and domain.

As a first iteration I suggest starting off with supporting the following:
* domain
* test_name
* probe_cc
* probe_asn

hellais:
This is still not implemented in: https://github.com/ooni/backend/blob/master/ooniapi/services/oonimeasurements/src/oonimeasurements/routers/v1/measurements.py

PR #929, opened 2025-01-27
Add support for passing a list to filters in list_measurements

hynnot:
This PR is about the issue https://github.com/ooni/backend/issues/795

Steps:
- Changes to accept the following filters as lists (domain, test_name, probe_cc, probe_asn)

Notes:
- Fix a problem with scores type.
- Fix a problem with the input variable.

### DIFF
--- a/ooniapi/services/oonimeasurements/tests/test_measurements.py
+++ b/ooniapi/services/oonimeasurements/tests/test_measurements.py
@@ -93,5 +93,6 @@ def test_list_measurements_with_multiple_values_to_filters_not_in_the_result(cli
     json = response.json()
     assert isinstance(json["results"], list), json
     assert len(json["results"]) > 0
+    domain_list = domainCollection.split(", ")
     for result in json["results"]:
-        assert any(domain in result["input"] for domain in domainCollection), result
+        assert any(domain in result["input"] for domain in domain_list), result
--- a/ooniapi/services/oonimeasurements/tests/test_measurements.py
+++ b/ooniapi/services/oonimeasurements/tests/test_measurements.py
@@ -1,0 +1,91 @@
+import pytest
+
+
+route = "api/v1/measurements"
+
+
+def test_list_measurements(client):
+    response = client.get(route)
+    json = response.json()
+
+    assert isinstance(json["results"], list), json
+    assert len(json["results"]) == 100
+
+
+def test_list_measurements_with_since_and_until(client):
+    params = {
+        "since": "2024-01-01",
+        "until": "2024-01-02",
+    }
+
+    response = client.get(route, params=params)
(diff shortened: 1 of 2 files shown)


ooniapi/services/oonimeasurements/tests/test_measurements.py
```diff
@@ -0,0 +1,91 @@
...
+    json = response.json()
+    assert isinstance(json["results"], list), json
+    assert len(json["results"]) > 0
+    for result in json["results"]:
+        assert result[filter_param] in filter_value, result
+
+
+def test_list_measurements_with_multiple_values_to_filters_not_in_the_result(client):
+    params = {
+        "domain": "cloudflare-dns.com, adblock.doh.mullvad.net, 1.1.1.1",
```
hellais, line 84: Isn't this not returning values because the domain parameters have extra whitespace after them?

I think it would be more clear to use a test that includes some domains that clearly do not exist in the result set.

hellais (changes requested):
Looks good to me. I think you can merge it when the issue with one of the tests is addressed.

ooniapi/services/oonimeasurements/tests/test_measurements.py
hynnot: The name of the test is the confusion here. We don't return the field domain in the result, this is why I can check the value (maybe this is an error and we should return it) About the whitespace it's not a problem because the split function removes them. 

About the second comment, I can add a domain not in our data because the result is the same. The list is an OR list.

hellais:
We realised in debugging this that `input` is none in the result listing:
```
             {'anomaly': False,
              'confirmed': False,
              'failure': False,
              'input': None,
              'measurement_start_time': '2021-07-09T19:14:20.000000Z',
              'measurement_uid': '20210709191422.444165_IT_webconnectivity_21a07ef7982fec55',
```

Which makes it not possible to test effectively the domain filtering. We should first investigate why it's None and then fix the domain filtering test.

ooniapi/services/oonimeasurements/tests/test_measurements.py
```diff
@@ -80,12 +83,15 @@ def test_list_measurements_with_multiple_values_to_filters(client, filter_param,
...
+        "domain": domainCollection
     }
 
     response = client.get(route, params=params)
 
     json = response.json()
     assert isinstance(json["results"], list), json
     assert len(json["results"]) > 0
+    for result in json["results"]:
+        assert any(domain in result["input"] for domain in domainCollection), result
```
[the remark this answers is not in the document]
hynnot: Good catch, I change in the next commit.

hellais (approved):
LGTM

PR #935, opened 2025-02-05
Dockerize Fastpath

hynnot:
This PR will add a docker compose configuration able to deploy the Fastpath.
- A **docker compose and dockerfile** are added to implement the docker deployment
  - This docker compose also includes a **clickhouse service** used for local development 
- **docker-related commands** are added to the makefile to simplify the development experience
- Some **broken tests** were marked with skip 
- The code is now compatible with the latest Python version

To deploy the docker container you can run: 
```
make docker
```

Note that we still use the `fastpath.conf` file for configuration, but I moved it to the root project dir and the dockerfile expects it to be there. The idea is that we will override this file through Ansible when deploying to EC2

Related to https://github.com/ooni/devops/issues/123

### DIFF
--- a/.github/workflows/build_deb_packages.yml
+++ b/.github/workflows/build_deb_packages.yml
@@ -3,7 +3,7 @@ on: pull_request
 
 jobs:
   build_deb_packages:
-    runs-on: "ubuntu-20.04"
+    runs-on: "ubuntu-24.04"
     container: debian:11
     steps:
       - name: Change Owner of Container Working Directory
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -3,7 +3,7 @@ on: push
 
 jobs:
   build_docs:
-    runs-on: "ubuntu-20.04"
+    runs-on: "ubuntu-24.04"
     steps:
       - name: Check out repository code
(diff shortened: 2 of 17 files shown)


fastpath/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Stage 1: Building
+FROM python:slim AS builder
```
hellais, line 2: I would suggest we be a bit more specific with the image names used for making the docker file.

Let's put in here explicitly `3.13.3-slim-bookworm` in case future versions of python or debian cause issues with running the fastpath.

fastpath/docker-compose.yml
```diff
@@ -0,0 +1,42 @@
+services:
+  fastpath:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: ooni-fastpath 
+    ports:
+      - "5000:5000"
```
hellais, line 8: What is listening on port 5000? I tried starting the docker container, but I get connection reset to port 5000, suggesting that perhaps there is no service listening there?

hellais (commented):
Thanks for working on this!

Left few comments and suggestions. Once solved we can merge.

Issue #925, opened 2025-01-22
Log ProbeUpdate metadata for anomaly detection


LDiazN:
This is a follow up issue for #923 

The Probe Update function in the old orchestra component logs useful metadata that we can use for anomaly detection. (see: https://github.com/ooni/orchestra/blob/master/registry/registry/handler/registry.go#L25)

If we assume that the update requests are not going to be tampered with as much as the measurement submit ones, we can use this to determine which inconsistencies are due to our geoip being off compared to the probe one, vs those which are due to the measurement metadata being inconsistent due to malice.

This can be done through the new update endpoint in the ooniprobe service, see #923, (https://github.com/ooni/backend/blob/dcf48d50fa283861aa24c125e182d27cf7662193/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py#L130)

PR #940, opened 2025-02-19
Log probe update metadata

LDiazN:
This PR adds Prometheus metrics to log probe update metadata that is usually ignored by the new update endpoint. 

The goal is to eventually use this data for anomaly detection in measurement update

This solves #925

### DIFF
--- a/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
@@ -193,8 +193,9 @@ def probe_update_post(probe_update: ProbeUpdate) -> ProbeUpdateResponse:
 
     # Info doesn't allows list, if we have a list we have to convert it
     # to string
-    if probe_update_dict["supported_tests"] is not None:
+    if probe_update_dict.get("supported_tests") is not None:
         tests = probe_update_dict["supported_tests"]
+        tests.sort()
         tests_str = ";".join(tests)
         probe_update_dict["supported_tests"] = tests_str
 
--- a/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
@@ -1,7 +1,7 @@
 import logging
 from datetime import datetime, timezone, timedelta
 import time
-from typing import List
+from typing import List, Optional
 
 from fastapi import APIRouter, Depends, HTTPException, Response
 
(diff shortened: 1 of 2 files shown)


ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -141,4 +174,17 @@ class ProbeUpdateResponse(BaseModel):
...
 def probe_update_post(probe_update: ProbeUpdate) -> ProbeUpdateResponse:
     log.info("update successful")
+    
+    # Log update metadata into prometheus
+    probe_update_dict = probe_update.model_dump(exclude_none=True)
+
+    # Info doesn't allows list, if we have a list we have to convert it 
+    # to string
+    if probe_update_dict['supported_tests'] is not None:
+        tests = probe_update_dict['supported_tests']
```
hellais, line 184: I suggest you sort the list before you build the string concatenation to make sure given the same set of tests it's always the same string irrespective of the order in which the probe sent them

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -141,4 +174,17 @@ class ProbeUpdateResponse(BaseModel):
 @router.put("/update/{client_id}", tags=["ooniprobe"])
 def probe_update_post(probe_update: ProbeUpdate) -> ProbeUpdateResponse:
     log.info("update successful")
+    
+    # Log update metadata into prometheus
+    probe_update_dict = probe_update.model_dump(exclude_none=True)
+
+    # Info doesn't allows list, if we have a list we have to convert it 
+    # to string
+    if probe_update_dict['supported_tests'] is not None:
```
hellais, line 183: Given that you have `exclude_none` above, won't this produce a key error if the probe didn't sent `supported_tests` or if it's set to `None`?

Maybe it's better to do `probe_update_dict.get('supported_tests')`?

LDiazN: Yes you're right, thanks for noticing!

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -141,4 +174,17 @@ class ProbeUpdateResponse(BaseModel):
...
+        tests = probe_update_dict['supported_tests']
+        tests_str = ";".join(tests)
+        probe_update_dict['supported_tests'] = tests_str
+
+    Metrics.PROBE_UPDATE_INFO.info(probe_update_dict)
```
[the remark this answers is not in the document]
LDiazN: I don't think it's going to be an issue with our current traffic volume to the `/update` endpoint. It's so low it doesn't even show up in Grafana, and if you check the logs in the old registry host you will see that the last time it was used was in `2025/01/07`. 

Also since most fields are strings with expected sizes, we can estimate how much space this would take, but to get an accurate estimation we would need to know the expected traffic once this is fixed (if it requires fixing), so my suggestion would be to just deploy it and check the dashboards and the Prometheus DB as you said

hellais (approved):
LGTM

PR #941, opened 2025-02-26
Add prom export rev proxy

### LINKED ISSUE #183: Use simhash to detect blockpages
Mentioned by @darkk in https://github.com/TheTorProject/ooni-pipeline/pull/62

LDiazN:
This PR will add [nginx prometheus exporter](https://github.com/nginx/nginx-prometheus-exporter) to the reverse proxy service to have metrics about the nginx service implementing the reverse proxy

I did this by installing the exporter in the same container as the reverse proxy and running it as a background process within the same container

Also note that requesting to `/metrics` to this server **is not forwarding to another host**, it will return metrics of the machine itself, not the proxied host

This PR closes https://github.com/ooni/devops/issues/183

### DIFF
--- a/ooniapi/services/oonirun/test.py
+++ b/ooniapi/services/oonirun/test.py
@@ -1,4 +1,0 @@
-import ooniprobe
-import ujson
-
-import ooniprobe.dist
--- a/ooniapi/services/oonirun/test.py
+++ b/ooniapi/services/oonirun/test.py
@@ -1,0 +1,4 @@
+import ooniprobe
+import ujson
+
+import ooniprobe.dist
--- a/ooniapi/services/reverseproxy/scripts/40-start-prom-exporter.sh
+++ b/ooniapi/services/reverseproxy/scripts/40-start-prom-exporter.sh
@@ -1,7 +1,0 @@
-#!/bin/bash
-
-# This script is used to start the prometheus exporter for nginx.
-# More about the exporter here: 
-# https://github.com/nginx/nginx-prometheus-exporter
-
-nginx-prometheus-exporter --nginx.scrape-uri=http://localhost:8080/stub_status &
--- a/ooniapi/services/reverseproxy/scripts/40-start-prom-exporter.sh
+++ b/ooniapi/services/reverseproxy/scripts/40-start-prom-exporter.sh
@@ -1,0 +1,7 @@
+#!/bin/bash
+
+# This script is used to start the prometheus exporter for nginx.
+# More about the exporter here: 
+# https://github.com/nginx/nginx-prometheus-exporter
+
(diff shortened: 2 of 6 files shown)


ooniapi/services/reverseproxy/scripts/40-start-prom-exporter.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+# This script is used to start the prometheus exporter for nginx.
+# More about the exporter here: 
+# https://github.com/nginx/nginx-prometheus-exporter
+
+nginx-prometheus-exporter --nginx.scrape-uri=http://localhost:8080/stub_status &
```
hellais, line 7: Is there the chance this process might crash and then it's not brought up again because it's run only using RUN in the dockerfile?

LDiazN: Suggestion applied, it worked well with `supervisord`

However we had to replace the original `docker-entrypoint.sh` with our own because changing the `CMD` of the docker file was preventing the original script from running the template set up required to create our configuration file, see: 

https://github.com/nginxinc/docker-nginx/blob/cffeb933620093bc0c08c0b28c3d5cbaec79d729/mainline/debian/docker-entrypoint.sh

ooniapi/services/oonirun/test.py
```diff
@@ -0,0 +1,4 @@
+import ooniprobe
+import ujson
+
+import ooniprobe.dist
```
hellais, line 4: What is this file?

ooniapi/services/reverseproxy/scripts/docker-entrypoint.sh
```diff
@@ -0,0 +1,50 @@
...
+# Copied from: 
+# https://github.com/nginxinc/docker-nginx/blob/cffeb933620093bc0c08c0b28c3d5cbaec79d729/mainline/debian/docker-entrypoint.sh
+
+entrypoint_log() {
+    if [ -z "${NGINX_ENTRYPOINT_QUIET_LOGS:-}" ]; then
+        echo "$@"
+    fi
+}
+
+if [ "$1" = "/usr/bin/supervisord" ] || [ "$1" = "nginx-debug" ]; then
```
hellais, line 15: Should you maybe also remove the `[ "$1" = "nginx-debug" ]` since you removed the `[ "$1" = "nginx" ]` line? Or maybe just add the supervisord to the check as an extra `||`?

LDiazN: Yeah I changed `nginx-debug` to `supervisord` so that it handles both calling styles

ooniapi/services/oonirun/test.py
LDiazN: Added by mistake, removed

PR #948, opened 2025-03-21
Port checkin endpoint

### LINKED ISSUE #942: Implement check-in endpoint inside of ooniprobe services component
We should port the check-in API endpoint over the new pattern so that we are able to collect metrics related to OONI Probe client metadata consistency for the anonymous credentials project

LDiazN:
The checkin endpoint provides useful metadata that we can use for the anonymous credential project. This port tries to have easier access to that data by porting it to ECS. 

## Relevant changes
- The GeoIP database is stored in `/var/lib/ooni/geoip` inside the container, it's downloaded everytime the container starts
- Some integration tests were not ported because they were no longer in use in the previous version
- Added a timer decorator function to measure function execution time
- The `ooni_download_geoip.py` script is now part of the ooniprobe service, it's used to download the database  

## Further work 
- Create a background task to update the geoip database
- Use our own geoip database from s3
- Check broken tests in the CI (docker-compose vs docker compose issue)

Closes #942


LDiazN:
closed with #951

PR #951, opened 2025-03-24
Bg geoip update

### LINKED ISSUE #950: Add periodic update of geoip database in the ooniprobe service
This is a follow up issue for: https://github.com/ooni/backend/pull/948

Now that the check-in endpoint, which requires the geoip database, is working on ECS, we need a way to update the database on disk periodically

We need to add a repeating background task that will update the database in the local filesystem (every hour maybe?)

LDiazN:
Adds support for running repeating tasks in background to the ooniprobe service using [fast-api utils](https://fastapi-utils.davidmontague.xyz/user-guide/repeated-tasks/)

Also add a new task that runs every hour that will check for updates and update the local geoip database within the service container. 

Another alternative to this approach is using the following library that does something similar: https://apscheduler.readthedocs.io/en/3.x/userguide.html

Closes #950


ooniapi/services/ooniprobe/tests/conftest.py
```diff
@@ -69,20 +71,55 @@ def client_with_bad_settings():
 
 
 JWT_ENCRYPTION_KEY = "super_secure"
+GEOIP_DB_DIR = f"{tempdir}/ooni/geoip"
```
hellais, line 74: You should define this path as a fixture and either make use of tmp_path of pytest (https://docs.pytest.org/en/stable/how-to/tmp_path.html), so that the tempfile is cleaned up after every test is run, or maybe even better cache it similarly to how this is done in pipeline (see: DATA_DIR in https://github.com/ooni/data/blob/main/oonipipeline/tests/conftest.py).

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -202,3 +233,362 @@ def probe_update_post(probe_update: ProbeUpdate) -> ProbeUpdateResponse:
...
+
+    conf["test_helpers"] = generate_test_helpers_conf()
+
+    resp["tests"] = {
+        "web_connectivity": {"urls": test_items},
+    }
+    resp["conf"] = conf
+    resp["utc_time"] = datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
+
+    test_names = (
```
hellais, line 398: Eventually we probably should place this list inside of the `common` so we don't have to make sure it's in sync in the multiple places where we use it.

ooniapi/services/ooniprobe/src/ooniprobe/main.py
```diff
@@ -66,8 +95,18 @@ class HealthStatus(BaseModel):
 async def health(
     settings=Depends(get_settings),
     db=Depends(get_postgresql_session),
+    clickhouse=Depends(get_clickhouse_session),
 ):
     errors = []
+    try:
+        query = """SELECT *
+        FROM fastpath FINAL
```
hellais, line 103: This query is incredibly heavy to run. I would not suggest running this as part of the health check as it will kill the database.

I would avoid placing any `FINAL` query in here and only place one that will return a subset of all the rows.

Maybe a reasonable query could be:
```
SELECT
COUNT()
FROM fastpath
WHERE measurement_start_time < NOW() AND measurement_start_time > NOW() - INTERVAL 3 HOUR
```

And check that you get some positive number in here.

ooniapi/services/ooniprobe/src/ooniprobe/download_geoip.py
```diff
@@ -0,0 +1,146 @@
...
+    )
+    GEOIP_UPDATED = metrics.Counter(
+        "ooni_geoip_updated", "How many times was the geoip database updated"
+    )
+    GEOIP_DOWNLOAD_TIME = metrics.Histogram(
+        "geoip_download_time", "How long it takes to download the DB"
+    )
+
+
+TS = datetime.now(timezone.utc).strftime("%Y-%m")
```
hellais, line 37: I think you have to place this TS calculation inside of the function body, otherwise it will be evaluated only upon module import which in the case of the period operation might be only once for the lifecycle of the application.

That is to say that it might not get updated to the next month, unless the the whole module gets re-imported, which I am not sure is the behavior of the period task scheduler.

ooniapi/services/ooniprobe/src/ooniprobe/download_geoip.py
```diff
@@ -0,0 +1,146 @@
...
+            shutil.copyfileobj(resp, out_file)
+    with gzip.open(str(tmp_gz_out)) as in_file:
+        with tmp_out.open("wb") as out_file:
+            shutil.copyfileobj(in_file, out_file)
+    tmp_gz_out.unlink()
+
+    try:
+        check_geoip_db(tmp_out)
+    except Exception as exc:
+        log.error(f"consistenty check on the geoip DB failed: {exc}")
```
hellais, line 115: Maybe delete the temporary outfile in here?

ooniapi/services/ooniprobe/src/ooniprobe/download_geoip.py
```diff
@@ -0,0 +1,146 @@
...
+
+    tmp_out.rename(db_dir / filename)
+    endtime = timeit.default_timer()  # End timer
+    Metrics.GEOIP_DOWNLOAD_TIME.observe(endtime - start_time)
+
+
+def update_geoip(db_dir: Path) -> None:
+    db_dir.mkdir(parents=True, exist_ok=True)
+    download_geoip(db_dir, ASN_URL, "asn.mmdb")
+    download_geoip(db_dir, CC_URL, "cc.mmdb")
```
hellais, line 127: The return codes of these two function calls are not getting checked, which means that if the consistency check fails, it will be considered to have been successfully downloaded while it isn't.

You should raise an exception for that case so that it's a hard fail

ooniapi/services/ooniprobe/tests/conftest.py
LDiazN: I notice in pipeline that the data is stored in a file within the project directory, but I don't see how it's being cleared afterwards. Shouldn't be that function decorated with a `@pytest.fixture(scope='session')` with a clean up code after the yield? Or is not clearing it  the intended usage to improve iteration speed?

ooniapi/services/ooniprobe/src/ooniprobe/main.py
LDiazN: I agree with this change. I made the original query based on this healthcheck from `oonimeasurements`: https://github.com/ooni/backend/blob/5dec72a5c4d47cebdd0b5969758efed5bd8da9b8/ooniapi/services/oonimeasurements/src/oonimeasurements/main.py#L81-L91

Should I change that one as well?

ooniapi/services/ooniprobe/src/ooniprobe/download_geoip.py
```diff
@@ -0,0 +1,146 @@
...
+log = logging.getLogger("ooni_download_geoip")
+
+log.addHandler(logging.StreamHandler(sys.stdout))
+log.setLevel(logging.DEBUG)
+
+
+def get_request(url):
+    req = Request(url)
+    # We need to set the user-agent otherwise db-ip gives us a 403
+    req.add_header("User-Agent", "ooni-downloader")
```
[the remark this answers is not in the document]
LDiazN: We can test this by deploying it to dev for a while and keeping an eye in the logs to see if we had problems downloading the DB

Also it's (normally) once a month, I don't think we might be blocked for traffic

ooniapi/services/ooniprobe/src/ooniprobe/download_geoip.py
LDiazN: I think `tmp_out` is renamed into the final file here: https://github.com/ooni/backend/blob/aae5dca3bdb679441fd522cf1812ec3a1385997b/ooniapi/services/ooniprobe/src/ooniprobe/download_geoip.py#L113  so I'm not sure if we can remove it

ooniapi/services/ooniprobe/tests/conftest.py
hellais: Yeah the idea is to not clear it so that if you run tests multiple times you don't have to redownload the assets. This also allows to cache that path in github actions for faster test runtime.

ooniapi/services/ooniprobe/src/ooniprobe/main.py
hellais: Yes we should change that one too. I must have missed that in previous reviews

PR #958, opened 2025-05-09
Add fix and test for priority calculation

hellais:
* Normalize the country_code to lowercase when matching the priority rules
* Add tests to prevent regression


LDiazN:
We have to apply it here as well: 
https://github.com/ooni/backend/blob/master/ooniapi/services/ooniprobe/src/ooniprobe/prio.py#L44

and port the test as well. This is the ported version we want to push as soon as the clickhouse migration is finished

PR #959, opened 2025-05-12
Update localhost refs

### LINKED ISSUE #954: Update all references to clickhouse://localhost with a configurable db_uri
In order to switch clickhouse over to the cluster we need to update all locations where we make use of clickhouse to a configurable URI.

So far we spotted these:
* https://github.com/ooni/backend/blob/d57a18feba8e93e4ab0b196df76d81e0abe95237/analysis/analysis/citizenlab_test_lists_updater.py#L104 
* https://github.com/ooni/backend/blob/d57a18feba8e93e4ab0b196df76d81e0abe95237/api/ooni_api_uploader.py#L55
* https://github.com/ooni/backend/blob/d57a18feba8e93e4ab0b196df76d81e0abe95237/analysis/analysis/analysis.py#L64

See: https://github.com/ooni/devops/issues/135

Maybe we can move the analysis tasks over to airflow while we are at it, see: https://github.com/ooni/backend/issues/926

These seem to not be used so we can ignore:
* https://github.com/ooni/backend/blob/d57a18feba8e93e4ab0b196df76d81e0abe95237/analysis/analysis/clickhouse_feeder.py#L52 
* https://github.com/ooni/backend/blob/d57a18feba8e93e4ab0b196df76d81e0abe95237/misc/tor_connectivity_monitor.py#L40

LDiazN:
Update all clickhouse localhost refs in backend to use a configuration parameter instead. 

It's set up such that if no such configuration is provided, it defaults to the old behaviour of having localhost hardcoded

The following config file requires an update:

### `analysis.conf`
```
# Managed by ansible, see roles/ooni-backend/tasks/main.yml
# [s3bucket]
# bucket_name = ooni-data-eu-fra-test
# aws_access_key_id =
# aws_secret_access_key =

[backup]
# space separated
table_names = citizenlab fastpath jsonl

# New
[DB]
uri = clickhouse url here
```

Closes #954 

Related to https://github.com/ooni/devops/pull/242


analysis/analysis/analysis.py
```diff
@@ -81,32 +83,58 @@ def parse_args() -> Namespace:
...
+    # Parse configs
+    conf_file = DEV_CONF_FILE if conf.devel else CONF_FILE
+
+    # nothing else to do if there's no config to parse
+    if not conf_file.exists():
+        conf.db_uri = DEFAULT_DB_URI
+        return     
+
+    cp = ConfigParser()
+    with conf_file.open("r") as f:
```
hellais, line 120: I would suggest dropping all this config file handling stuff and only allow setting db-uri via the CLI flag `--db-uri`.

hellais (changes requested):
Only support setting db-uri using CLI flag

analysis/analysis/analysis.py
hellais: The reason is that we are potentially adding some extra code paths that can be tricky to debug if something is not working as intended.

LDiazN: How do we set up the flag `--db-uri`  when deployed?  I originally used the conf file because it's where we specify configurations in Ansible

analysis/analysis/analysis.py
```diff
@@ -61,7 +61,9 @@
 
 log = logging.getLogger("analysis")
 metrics = setup_metrics(name="analysis")
-DB_URI = "clickhouse://api:api@localhost/default"
+DEFAULT_DB_URI = "clickhouse://api:api@localhost/default"
+CONF_FILE = Path("/etc/ooni/analysis.conf")
```
hellais, line 65: Are these used anywhere? Maybe they can be removed?

hellais (approved):
I would remove the potentially unused variable and then it's good to go.

analysis/analysis/analysis.py
LDiazN: DEFAULT_DB here: https://github.com/ooni/backend/blob/c35ce2516b98b653cc994a6091bdd1eef65f7ac7/analysis/analysis/analysis.py#L86
 
 CONF_FILE is not, removing it

Issue #955, opened 2025-05-05
Implement initial OONI Run v2.1


hellais:
This is about implementing a first iteration of OONI Run v2.1 with support for dynamic URL listing based on the spec in: https://github.com/ooni/spec/pull/301

aagbsn:
Marking this as done as it is deployed on prod; can we make #916 a v2.2 feature?

PR #962, opened 2025-05-23
Oonirun v2 1

LDiazN:
Add DB model migrations, new view models and some tests. Related to #955

### DIFF
--- a/ooniapi/common/src/common/alembic/versions/b860eb79750f_add_targets_name_and_inputs_extra_.py
+++ b/ooniapi/common/src/common/alembic/versions/b860eb79750f_add_targets_name_and_inputs_extra_.py
@@ -1,0 +1,32 @@
+"""Add targets_name and inputs_extra columns to oonirun_nettest
+
+Revision ID: b860eb79750f
+Revises: 8e7ecea5c2f5
+Create Date: 2025-05-21 15:44:32.959349
+
+"""
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision: str = 'b860eb79750f'
+down_revision: Union[str, None] = '8e7ecea5c2f5'
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    op.add_column("oonirun_nettest", 
+        sa.Column("targets_name", sa.String(), nullable=True)
+    )
+    op.add_column("oonirun_nettest",
+        sa.Column("inputs_extra", sa.JSON(), nullable=True)
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("oonirun_nettest", "targets_name")
+    op.drop_column("oonirun_nettest", "inputs_extra")
--- a/ooniapi/services/oonirun/src/oonirun/routers/v2.py
+++ b/ooniapi/services/oonirun/src/oonirun/routers/v2.py
@@ -5,7 +5,7 @@ https://github.com/ooni/spec/blob/master/backends/bk-005-ooni-run-v2.md
 """
 
 from datetime import datetime, timedelta, timezone
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple, Any
 import logging
 
 import sqlalchemy as sa
@@ -113,6 +113,16 @@ class OONIRunLinkBase(BaseModel):
     expiration_date: datetime = Field(
         default_factory=lambda: utcnow_seconds() + timedelta(days=30 * 6),
         description="future time after which the ooni run link will be considered expired and no longer editable or usable (defaults to 6 months from now)",
+    )
+
+    targets_name : Optional[str] = Field (
+        default=None, 
+        description="string used to specify during creation that the input list should be dynamically generated."
+    )
+
+    inputs_extra : Optional[List[Dict[str, Any]]] = Field (
+        default = None,
+        description = "provides a richer JSON array containing extra parameters for each input. If provided, the length of inputs_extra should match the length of inputs."
     )
 
 
--- a/ooniapi/services/oonirun/src/oonirun/routers/v2.py
+++ b/ooniapi/services/oonirun/src/oonirun/routers/v2.py
@@ -51,6 +51,15 @@ class OONIRunLinkNettest(BaseModel):
         default=False, title="if this test should be enabled by default for manual runs"
     )
 
+    targets_name : Optional[str] = Field (
+        default=None, 
+        description="string used to specify during creation that the input list should be dynamically generated."
+    )
+
+    inputs_extra : Optional[List[Dict[str, Any]]] = Field (
+        default = None,
+        description = "provides a richer JSON array containing extra parameters for each input. If provided, the length of inputs_extra should match the length of inputs."
+    )
 
 class OONIRunLinkEngineDescriptor(BaseModel):
     revision: str = Field(title="revision of the nettest descriptor")
@@ -114,17 +123,6 @@ class OONIRunLinkBase(BaseModel):
         default_factory=lambda: utcnow_seconds() + timedelta(days=30 * 6),
         description="future time after which the ooni run link will be considered expired and no longer editable or usable (defaults to 6 months from now)",
     )
-
-    targets_name : Optional[str] = Field (
-        default=None, 
-        description="string used to specify during creation that the input list should be dynamically generated."
-    )
-
-    inputs_extra : Optional[List[Dict[str, Any]]] = Field (
-        default = None,
-        description = "provides a richer JSON array containing extra parameters for each input. If provided, the length of inputs_extra should match the length of inputs."
-    )
-
 
 class OONIRunLink(OONIRunLinkBase):
     oonirun_link_id: str
--- a/ooniapi/services/oonirun/src/oonirun/routers/v2.py
+++ b/ooniapi/services/oonirun/src/oonirun/routers/v2.py
@@ -42,6 +42,7 @@ class OONIRunLinkNettest(BaseModel):
         default=[], title="list of input dictionaries for the nettest"
     )
     options: Dict = Field(default={}, title="options for the nettest")
+    # TODO(luis): Not in the new spec. Should be removed?
     backend_options: Dict = Field(default={}, title="options to send to the backend")
     is_background_run_enabled_default: bool = Field(
         default=False,
@@ -131,6 +132,11 @@ class OONIRunLinkBase(BaseModel):
     )
 
 class OONIRunLink(OONIRunLinkBase):
+
+    # TODO(luis): spec mentions that there should be a title, description and author here.
+    # see: https://github.com/ooni/spec/blob/oonirun-v2.1/backends/bk-005-ooni-run-v2.md#response-body
+    # should we add it?
+
     oonirun_link_id: str
     date_created: datetime = Field(
         description="time when the ooni run link was created"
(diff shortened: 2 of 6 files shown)


ooniapi/services/oonirun/src/oonirun/routers/v2.py
```diff
@@ -151,9 +171,9 @@ class OONIRunLinkCreateEdit(OONIRunLinkBase):
 )
 def create_oonirun_link(
     create_request: OONIRunLinkCreateEdit,
+    db: PostgresSession,
```
hellais, line 174: This is an interesting way to defined the dependencies I wasn't aware of: https://fastapi.tiangolo.com/tutorial/dependencies/#declare-the-dependency-in-the-dependant.

I wonder if at some point we should refactor the other dependencies to use it too.

ooniapi/services/oonirun/src/oonirun/routers/v2.py
```diff
@@ -41,6 +41,7 @@ class OONIRunLinkNettest(BaseModel):
     inputs: List[str] = Field(
         default=[], title="list of input dictionaries for the nettest"
     )
+    # TODO(luis): Options and backend_options not in the new spec. Should be removed?
```
hellais, line 44: We should for sure drop `backend_options`. I am not sure if it's useful to have a top level `options` that applies to all the inputs inside of the `nettest` or if it's redundant as we can include it directly inside of the `inputs_extra`.

@DecFox what do you think?

ooniapi/services/oonirun/src/oonirun/routers/v2.py
```diff
@@ -51,6 +52,24 @@ class OONIRunLinkNettest(BaseModel):
         default=False, title="if this test should be enabled by default for manual runs"
     )
 
+    targets_name: Optional[str] = Field(
```
hellais, line 55: I think we should do some validation on create and edit, that if `targets_name` is provided, you can't also provide `input` and `inputs_extra`, since these are to be populated based on the value of `targets_name`.

LDiazN: is it an error to not provide any of them?

hellais: It's not an error, because you can have a test that has no input for it (eg. `whatsapp`)

ooniapi/services/oonirun/src/oonirun/routers/v2.py
```diff
@@ -33,16 +37,26 @@
...
+    def probe_asn_int(self) -> int:
+        return int(self.probe_asn.replace("AS", ""))
 
 class OONIRunLinkNettest(BaseModel):
     test_name: str = Field(
         default="", title="name of the ooni nettest", min_length=2, max_length=100
     )
-    inputs: List[str] = Field(
+    inputs: Optional[List[str]] = Field(
         default=[], title="list of input dictionaries for the nettest"
```
hellais, line 56: I think `inputs` should be initialized to `None`, otherwise the validation for `targets_name` not being preset at the same time as `inputs` will fail.

hellais: We can otherwise change this to be the empty list, but then we should change that also for `inputs_extra`

ooniapi/services/oonirun/src/oonirun/routers/v2.py
```diff
@@ -334,22 +397,68 @@ def edit_oonirun_link(
...
+
     date_created = oonirun_link.nettests[0].date_created
     nettests = []
     for nt in oonirun_link.nettests:
         if revision and nt.revision != revision:
             continue
         date_created = nt.date_created
+        inputs, inputs_extra = nt.inputs, nt.inputs_extra
+        targets_name = nt.targets_name
+        if nt.targets_name is not None and meta is not None:
```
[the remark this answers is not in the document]
LDiazN: The thing is that this function is also called when no dynamic test list is required. If you don't pass the meta it just gives you whatever is in the `inputs` field in the DB (empty for nettests with dynamic test lists)

If we require the meta here we have to: 
1. Option 1: Always compute a dynamic list when this function is called and  and making the meta  required for endpoints  that don't currently use it. Ex: `/v2/oonirun/links` that lists descriptors, `/v2/oonirun/links/{oonirun_link_id}/full-descriptor/{revision_number}` to get a descriptor 

2. Option 2: We make a new function to get the nettest from the DB without the dynamic test list

I would suggest to split the dynamic test list calculation and fetching from the DB into two different functions, since these are two different tasks, and only compute the dynamic test lists explicitly when needed

ooniapi/services/oonirun/src/oonirun/routers/v2.py
```diff
@@ -434,8 +544,28 @@ def get_oonirun_link_revisions(
...
 
+class XOoniNetworkInfo(BaseModel):
+    probe_asn : str
+    probe_cc : str
+    network_type : str
 
-@router.get(
+    @staticmethod
+    def get_header_pattern() -> str:
+        return r'^([a-zA-Z0-9]+),([a-zA-Z0-9]+) \(([a-zA-Z0-9]+)\)$'
```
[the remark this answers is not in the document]
LDiazN: I agree with this, I will still keep the regex since it's used for validation but a simpler format would be better (we also have to update the spec to reflect this)

ooniapi/services/oonirun/src/oonirun/routers/v2.py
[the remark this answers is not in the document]
hellais: Doesn't `nt.targets_name is not None` imply that it's getting called as a dynamic test list and hence `meta` is required?

If this function is called when it's not dynamic, shouldn't `nt.targets_name` be none and therefore the assertion will not be triggered?

ooniapi/services/oonirun/src/oonirun/routers/v2.py
```diff
@@ -559,8 +708,11 @@ def list_oonirun_links(
         Optional[bool],
         Query(description="List also expired descriptors"),
     ] = None,
+    only_latest: Annotated[
```
hellais, line 698: Let's drop this parameter from this PR.

ooniapi/services/oonirun/src/oonirun/routers/v2.py
```diff
@@ -33,16 +37,26 @@
 def utcnow_seconds():
     return datetime.now(timezone.utc).replace(microsecond=0)
 
+class OonirunMeta(BaseModel):
+    run_type : str = Field(description="Run type", pattern="^(timed|manual)$")
+    is_charging : bool = Field(description="If the probe is charging")
+    probe_asn : str = Field(pattern=r"^([a-zA-Z0-9]+)$")
```
hellais, line 43: This pattern seems a bit too lax. I would say it should be: `^(AS)?([0-9]{1,10})$`

ooniapi/services/oonirun/src/oonirun/routers/v2.py
```diff
@@ -33,16 +37,26 @@
...
+    probe_cc : str = Field(description="Country code. Ex: VE")
+    network_type : str = Field(description="Ex: wifi")
```
hellais, line 45: We also have an explicit list of valid `network_type`. @aanorbel @sdsantos can you provide us this list so we can add it here to the validation list?

ooniapi/services/oonirun/tests/integ/test_dynamic_lists.py
```diff
@@ -0,0 +1,188 @@
...
+        "cc": "*",
+        "domain": "ooni.org",
+        "priority": 99999,
+        "url": "*",
+        "sign" : 1
+    } 
+    query = "INSERT INTO url_priorities (sign, category_code, cc, domain, url, priority) VALUES"
+    insert_click(clickhouse_db, query, [values])
+    yield 
+    clickhouse_db.execute("DELETE FROM url_priorities WHERE domain='www.ooni.com'")
```
hellais, line 167: Shouldn't domain here be `ooni.org`?

ooniapi/services/oonirun/tests/test_oonirun.py
```diff
@@ -585,6 +606,251 @@ def test_oonirun_revisions(client, client_with_user_role):
...
+    r = client_with_user_role.post(f"/api/v2/oonirun/links/{j['oonirun_link_id']}/engine-descriptor/latest", json=SAMPLE_META)
+    assert r.status_code == 200, r.json()
+
+    j = r.json()
+    assert j['nettests'][0]['targets_name'] == "websites_list_prioritized" 
+
+    # TODO(luis) Finish this test
+
+
+# TODO(luis) finish this test for checking the parsing of user agent headers
```
hellais, line 817: Are you planning to address these TODO in this PR or as following ones? If the later, then it would be a good idea to make an issue for them.

hellais (changes requested):
I left some comments for things to go over before merging.

Also, I noticed in several places there is some odd formatting (eg. there being an extra space around `:`, like `"foo" : "bar"`). Could you run `black` on all the files you edited to make the formatting consistent?

ooniapi/services/oonirun/tests/test_oonirun.py
LDiazN: I think we can mark it as complete, because I put that before doing integration tests and the simpler header format so I don't think there's something else to add in this tests

ooniapi/services/oonirun/src/oonirun/routers/v2.py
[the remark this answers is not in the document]
LDiazN: I wasn't aware of this kind of attack, thanks for sharing it!
I will change the parsing to use split by "," then

ooniapi/common/src/common/alembic/versions/b860eb79750f_add_targets_name_and_inputs_extra_.py
```diff
@@ -0,0 +1,32 @@
...
+        sa.Column("targets_name", sa.String(), nullable=True)
+    )
+    op.add_column("oonirun_nettest",
+        sa.Column("inputs_extra", sa.ARRAY(sa.JSON()), nullable=True)
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("oonirun_nettest", "targets_name")
+    op.drop_column("oonirun_nettest", "inputs_extra")
```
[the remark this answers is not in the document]
hellais: Older probes aren't using it and we have anyways dropped it from the JSON API format.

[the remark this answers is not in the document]
LDiazN: Sure, then I will drop the `backend` and `backend_options` columns and remove it from the API

ooniapi/services/oonirun/src/oonirun/routers/v2.py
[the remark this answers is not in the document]
hellais: After discussing this on slack, I think now I have a better understanding of the issue. I like the solution you proposed in the last commit of splitting out the two functions.

For me that's a good way to to do it 👍

ooniapi/services/oonirun/src/oonirun/routers/v2.py
hellais: So in thinking more about the topic of the `options` key, I don't think we can get rid of it. It's actually not redundant with `inputs_extra` since you would not be able to pass the options for a test using `inputs_extra` if that test takes no inputs.

As a result of this, I think we might have to keep `options`

hellais (approved):
This is PR is good to go.

The steps to get this into production are:
* [ ] Deploy this branch into the dev environment by updating the terraform configs
* [ ] Test that the dev environment with OONI Run v2.1 works on the OONI Probe mobile apps
* [ ] Test that the dev environment with OONI Run v2.1 works with `miniooni`
* [ ] Merge this PR which will lead to a production deployment

PR #971, opened 2025-07-14
Fastpath writes to disk

### LINKED ISSUE #968: Add support for writing files to disk in fastpath
While reviewing https://github.com/ooni/devops/pull/240 with @LDiazN we realised we are still missing a piece to get this deployed.

Namely, at the moment files are written to disk by the OONI API and not the fastpath (see: https://github.com/ooni/backend/blob/master/api/ooniapi/probe_services.py#L859). This means that we would have to somehow share the disk between fastpath and OONI API, but the API we would like it to live on ECS.

Instead of sharing the disk we concluded that the best way forward is to move the code for writing to disk into the fastpath.

The data payload is already being transfered over the network: https://github.com/ooni/backend/blob/master/api/ooniapi/probe_services.py#L945C22-L945C26.

We should be sure to log several metrics related to this, especially when the upload fails: https://github.com/ooni/backend/blob/master/api/ooniapi/probe_services.py#L949 and upload the unsubmitted file to s3 at a particular prefix.

And then add logic to write the json blob to disk in the `msm_processor` https://github.com/ooni/backend/blob/master/fastpath/fastpath/core.py#L1579. Be careful to not do it in the process_measurement function, as that's called also by the reprocessor and we don't want to duplicate files when we are reprocessing them.

LDiazN:
With our current setup the API in the monolith is writing report data to disk with the `/report/{id}` endpoint while also sending the same data to the fastpath. 

This data is meant to be processed by the api uploader.  

However since we intend to move this setup to AWS, we need a different approach, see #968 

This PR will move the file-writing part to the Fastpath once the measurement is received. This is necessary because the ooniprobe service is unable to share disk with the EC2 machine running fastpath and api uploader

The file-writing feature is under a feature flag and deactivated by default. To activate it we add this in `fastpath.conf`
```
[DEFAULT]
...
write_to_disk = true
```

It will also add the config parameter `msmt_spool_dir` to the configuration file. It will only be parsed when `write_to_disk == true`

depends on #969 
closes #968

### DIFF
--- a/ooniapi/services/ooniprobe/src/ooniprobe/prio.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/prio.py
@@ -65,7 +65,7 @@ def compute_priorities(entries: tuple, prio_rules: tuple) -> list:
     return sorted(test_list, key=lambda k: k["weight"], reverse=True)
 
 
-@timer
+@timer(name="fetch_reactive_url_list")
 def fetch_reactive_url_list(
     clickhouse_db: Clickhouse, cc: str, probe_asn: int
 ) -> tuple:
@@ -105,7 +105,7 @@ def fetch_reactive_url_list(
     return tuple(r)
 
 
-@timer
+@timer(name="fetch_prioritization_rules")
 def fetch_prioritization_rules(clickhouse_db: Clickhouse, cc: str) -> tuple:
     sql = """SELECT category_code, cc, domain, url, priority
     FROM url_priorities WHERE cc = :cc OR cc = '*' OR cc = ''
@@ -114,7 +114,7 @@ def fetch_prioritization_rules(clickhouse_db: Clickhouse, cc: str) -> tuple:
     return tuple(q)
 
 
-@timer
+@timer(name="generate_test_list")
 def generate_test_list(
     clickhouse: Clickhouse,
     country_code: str,


fastpath/fastpath/core.py
```diff
@@ -1601,6 +1613,39 @@ def flag_measurements_with_wrong_date(msm: dict, msmt_uid: str, scores: dict) ->
...
         scores["msg"] = "Measurement start time too old"
 
+def write_measurement_to_disk(msm_tup) -> None:
+    """Write this measurement to disk so that it can be 
+    processed by the measurement uploader
+
+    Args:
+        msm_tup: Measurement tuple as it comes from the work queue
+    """
+    if not conf.write_to_disk:
```
hellais, line 1623: I think it's maybe more clear to tell this is a `noop` if we place this check inside of the function that's calling it, that is `msm_processor`.

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -0,0 +1,256 @@
...
+    h = sha512(data).hexdigest()[:16]
+    ts = now.strftime("%Y%m%d%H%M%S.%f")
+
+    # msmt_uid is a unique id based on upload time, cc, testname and hash
+    msmt_uid = f"{ts}_{cc}_{test_name}_{h}"
+    Metrics.MSMNT_RECEIVED_CNT.inc()
+
+    compare_probe_msmt_cc_asn(cc, asn, request, cc_reader, asn_reader)
+    N_RETRIES = 3
+    for t in range(N_RETRIES):
```
hellais, line 195: If we are going to be doing some retries, we should place some kind of jitter in the retries to avoid a thundering herd from the workers. 

I would suggest we do an exponential backoff with jitter using something like this:
```
sleep_amount = random.uniform(0, min(3, 0.3 * 2 ** t))
await asyncio.sleep(sleep_amount)
```

This should give us an exponential backoff with some jitter up to 3 seconds maximum.

hellais (changes requested):
Thanks for putting this together it looks very good. I have left a few comments for things to change. Once addressed we can move it forward.

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -0,0 +1,276 @@
...
+    Metrics.MSMNT_RECEIVED_CNT.inc()
+
+    compare_probe_msmt_cc_asn(cc, asn, request, cc_reader, asn_reader)
+    # Use exponential back off with jitter between retries
+    N_RETRIES = 3
+    for t in range(N_RETRIES):
+        try:
+            url = f"{settings.fastpath_url}/{msmt_uid}"
+            # TODO would be nice to make this async due to the size of the data being transmitted.
+            urlopen(url, data, 59)
```
hellais, line 213: I think we might have to make this async otherwise we will have issues with these requests blocking the API thread.

LDiazN: Agree, I used httpx as async client since it was already installed in ooniprobe

Issue #965, opened 2025-05-30
Investigate why fingerprint is not getting detected in Indonesia


hellais:
In the following measurement we would have expected the fingerprint to have been detected, but it isn't: https://explorer.ooni.org/m/20250527233111.802678_ID_webconnectivity_5a37ec909dbf14e3.

It should have triggered this fingerprint: https://github.com/ooni/blocking-fingerprints/blob/main/fingerprints_http.csv#L1707

We should add a test for the scoring to see why it's not detecting it: https://github.com/ooni/backend/blob/master/fastpath/fastpath/tests/test_functional.py#L679.

hellais:
As it turns out this affects all data since commit https://github.com/ooni/backend/commit/ec201fc993534d88ad61ac0df20f3a02d91d07c6 (March 2023) and all fingerprints that are detectable through HTTP headers.

PR #977, opened 2025-07-28
Fix off-by-one in HTTP fingerprint matching

hellais:
Closes: https://github.com/ooni/backend/issues/965


hellais:
This bug was introduced in https://github.com/ooni/backend/commit/ec201fc993534d88ad61ac0df20f3a02d91d07c6

PR #978, opened 2025-08-05
Fix bad failure field

LDiazN:
Quick fix for validation errors for measurement meta:

- We were using `failure` instead of `msm_failure` to retrieve the failure field from the clickhouse query
- We were passing the ASN as int rather than a string, as expected by the `MeasurementMeta` model

### DIFF
--- a/ooniapi/services/oonimeasurements/tests/test_measurements.py
+++ b/ooniapi/services/oonimeasurements/tests/test_measurements.py
@@ -1,5 +1,8 @@
 import pytest
-
+from clickhouse_driver import Client as Clickhouse
+from oonimeasurements.common.clickhouse_utils import query_click_one_row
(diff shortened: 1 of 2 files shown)


ooniapi/services/oonimeasurements/tests/test_measurements.py
```diff
@@ -96,3 +99,20 @@ def test_list_measurements_with_multiple_values_to_filters_not_in_the_result(cli
     domain_list = domainCollection.split(", ")
     for result in json["results"]:
         assert any(domain in result["input"] for domain in domain_list), result
+
+def test_failure_format(db):
+    ch = Clickhouse.from_url(db)
+    msm = query_click_one_row(ch, "select * from fastpath where test_name = 'web_connectivity'", {}) or {}
```
hellais, line 105: It's probably better to add a `LIMIT 1` in here as well.

PR #983, opened 2025-08-07
Oonimsm fixes

LDiazN:
Small fixes to make the oonimeasurements and fastpath components integrate properly:

- Move the protocol specification to the `other_collectors` setting instead of enforcing it in oonimeasurements
- Fix an issue where the fastpath would rewrite the ID of the measurement when writing to disk, generating slightly different IDs to the ones in the DB and leading to 404 errors. This doesn't affect the monolith version, only the EC2 version. 

Related to https://github.com/ooni/devops/pull/262

### DIFF
--- a/fastpath/fastpath/core.py
+++ b/fastpath/fastpath/core.py
@@ -1622,6 +1622,7 @@ def write_measurement_to_disk(msm_tup) -> None:
         msm_tup: Measurement tuple as it comes from the work queue
     """
 
+    # msmt_uid is a unique id based on upload time, cc, testname and hash
     data, measurement, msmt_uid = msm_tup
     ts, cc, test_name, h = msmt_uid.split("_")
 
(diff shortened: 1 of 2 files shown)


fastpath/fastpath/core.py
```diff
@@ -1632,15 +1633,9 @@ def write_measurement_to_disk(msm_tup) -> None:
     msmtdir = spooldir / "incoming" / dirname
     msmtdir.mkdir(parents=True, exist_ok=True)
 
-    h = sha512(data).hexdigest()[:16]
-    ts = now.strftime("%Y%m%d%H%M%S.%f")
```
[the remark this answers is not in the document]
LDiazN: Good catch! 
I tested it but it was working well because most of the time measurements came at the same hour to the fastpath and ooniprobe, but might have crashed on inter-hour requests or if the machine takes too long to process measurements in the queue

fastpath/fastpath/core.py
```diff
@@ -1622,25 +1622,20 @@ def write_measurement_to_disk(msm_tup) -> None:
         msm_tup: Measurement tuple as it comes from the work queue
     """
 
+    # msmt_uid is a unique id based on upload time, cc, testname and hash
     data, measurement, msmt_uid = msm_tup
     ts, cc, test_name, h = msmt_uid.split("_")
 
-    now = datetime.now(timezone.utc)
+    now = datetime.strptime(ts, "%Y%m%d%H%M%S.%f")
```
hellais, line 1629: I think you can just extract the `hour` value directly from the `ts` string by extracting the first 8 characters, there is no need to go through the datetime parser.

I suggest doing that to keep it simpler.

LDiazN: Wouldn't that be the first 10 characters? 
Example: 
```
>>> now.strftime("%Y%m%d%H")
'2025080813'
>>> len(now.strftime("%Y%m%d%H"))
10
```

Issue #943, opened 2025-03-13
Iterate on the observations and experiment results API endpoints for explorer integration


hellais:
Placeholder for any work related to the backend API

hellais:
TODO:
* [ ] Map in observation aggregation endpoint all unknown failures to the same unknown_failure string
* [ ] Look into the performance of the observation aggregation query

PR #984, opened 2025-08-07
Improve observations api

hellais:
Adds support for
* Consolidate unknown_failure into the same failure string
* Remove `after_CH` failure string

xref: https://github.com/ooni/backend/issues/943

### DIFF
--- a/ooniapi/services/oonimeasurements/src/oonimeasurements/routers/data/aggregate_observations.py
+++ b/ooniapi/services/oonimeasurements/src/oonimeasurements/routers/data/aggregate_observations.py
@@ -124,13 +124,13 @@ async def get_aggregation_observations(
         columns.append(
             f"""multiIf(
     dns_failure IS NOT NULL,
-    CONCAT('dns_', dns_failure),
+    CONCAT('dns_', IF(startsWith(dns_failure, 'unknown_failure'), 'unknown_failure', dns_failure)),
     tcp_failure IS NOT NULL,
-    CONCAT('tcp_', tcp_failure),
+    CONCAT('tcp_', IF(startsWith(tcp_failure, 'unknown_failure'), 'unknown_failure', tcp_failure)),
     IF(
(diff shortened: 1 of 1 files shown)


DecFox (approved):

ooniapi/services/oonimeasurements/src/oonimeasurements/routers/data/aggregate_observations.py
```diff
@@ -123,17 +123,15 @@ async def get_aggregation_observations(
...
-tcp_failure IS NOT NULL,
-CONCAT('tcp_', tcp_failure),
-multiIf(
-tls_handshake_last_operation = 'write_1',
-CONCAT('tls_', tls_failure, '_after_CH'),
-tls_failure IS NULL AND tls_server_name IS NOT NULL,
-'none',
-CONCAT('tls_', tls_failure)
-)
+    dns_failure IS NOT NULL,
```
LDiazN, line 126: This is applied over tcp, tls and dns failure. Is it necessary to do it for `http_failure` as well?

Otherwise LGTM!

hellais: We exclude by design `http` related observations from these counts. The reason to do that is that we don't know for the http(s) observations if the failure occurred at `tcp` or `tls` or `http` or `https` for certain older measurements which didn't properly annotate the address field.

It's mentioned a bit below in a comment that's maybe a bit too cryptic. I am going to push a new commit that adds a comments that a bit more explicit and better explains this.

Thanks for pointing out that it's not so clear why we are doing this and it's happening by design.

hellais: In the end I opted to actually include the `http_` and `https_` failures which can be mapped to an IP. We still exclude  those which cannot be accounted for precisely, but this way we are able to know which failures were uniquely occurring at https or http, but were succeeding at `tls` or `tcp`

PR #987, opened 2025-08-13
Fix broken debian packages CI

LDiazN:


hellais:
superseeded by latest PR

PR #988, opened 2025-08-14
Set fastpath user in docker

LDiazN:
Add the fastpath used in docker with a fixed user

Also create its expected files with the right access permissions

See: https://github.com/ooni/devops/issues/265
Related to: https://github.com/ooni/devops/pull/262


hellais (approved):
LGTM!

PR #992, opened 2025-08-27
Fix bad raw_measurement response encoding

LDiazN:
the `/raw_measurement` endpoint in `oonimeasurements` is returning a string with a json instead of a json itself 

This happened because the measurement body is downloaded as a string (a json string), and it's converted into a json again by FastApi's `JSONResponse`:

Our code: 
https://github.com/ooni/backend/blob/b04cfe72ad28fb68602756bc9300ae225b35e1cc/ooniapi/services/oonimeasurements/src/oonimeasurements/routers/v1/measurements.py#L407C1-L409C20

`JSONResponse`: 
https://github.com/encode/starlette/blob/cb8f84f5284dc301ca7a31eb732b9e140769dd48/starlette/responses.py#L192C4-L199C4


hellais (approved):
LGTM

Issue #1008, opened 2025-09-17
oonimeasurements probe_asn filter does not work as expected when omitting "AS" prefix


hellais:
In the previous version of the list_measurements endpoint we would support passing the `probe_asn` filter in both the format `probe_asn=AS123` and `probe_asn=1234`.

The current version of the API only works when it's passed **with** the AS prefix, leading to explorer not listing measurements when applying the probe_asn filter.

To reproduce:
```
curl -s 'https://api.ooni.org/api/v1/measurements?limit=50&probe_cc=US&domain=x.com&probe_asn=174&test_name=web_connectivity&since=2024-08-09&until=2024-08-10&order=desc' | jq '.results | length'
0
```

while with the `AS` prefix:
```
curl -s 'https://api.ooni.org/api/v1/measurements?limit=50&probe_cc=US&domain=x.com&probe_asn=AS174&test_name=web_connectivity&since=2024-08-09&until=2024-08-10&order=desc' | jq '.results | length'
28
```

These are the results for the previous version of the API, which can be reproduced by logging into backend-fsn.ooni.org:
```
$ curl -s 'http://localhost:8000/api/v1/measurements?limit=50&probe_cc=US&domain=x.com&probe_asn=174&test_name=web_connectivity&since=2024-08-09&until=2024-08-10&order=desc' | jq '.results | length'
28
$ curl -s 'http://localhost:8000/api/v1/measurements?limit=50&probe_cc=US&domain=x.com&probe_asn=AS174&test_name=web_connectivity&since=2024-08-09&until=2024-08-10&order=desc' | jq '.results | length'
28
```

hellais:
The issue is that it's appending to the list in order to apply the filter only when it starts with the `AS` prefix: https://github.com/ooni/backend/blob/master/ooniapi/services/oonimeasurements/src/oonimeasurements/routers/v1/measurements.py#L747

PR #1009, opened 2025-09-17
Fix bug with filtering by probe_asn without the AS prefix

hellais:
Closes: https://github.com/ooni/backend/issues/1008

### DIFF
--- a/ooniapi/services/oonimeasurements/tests/test_measurements.py
+++ b/ooniapi/services/oonimeasurements/tests/test_measurements.py
@@ -173,6 +173,7 @@ def test_raw_measurement_args_optional(client, monkeypatch, maybe_download_fixtu
     resp = client.get("/api/v1/raw_measurement", params={})
     assert resp.status_code == 400, resp.status_code
 
+
 def test_raw_measurement_returns_json(client, monkeypatch, maybe_download_fixtures):
     """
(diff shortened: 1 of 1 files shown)


ooniapi/services/oonimeasurements/tests/test_measurements.py
```diff
@@ -37,6 +37,7 @@ def test_list_measurements_with_since_and_until(client):
         ("test_name", "web_connectivity"),
         ("probe_cc", "IT"),
         ("probe_asn", "AS30722"),
+        ("probe_asn", "30722"),
```
LDiazN, line 40: This change is breaking the test in line 55 bc the api always returns the ASN formated as `AS<number>` , you can fix it by formatting `filter_value` before line 55 like: 
```
    if filter_param == "probe_asn":
        filter_value = filter_value if filter_value.startswith("AS") else "AS" + filter_value
```

LDiazN: There's a similar issue with this test: 
https://github.com/ooni/backend/pull/1009/files#diff-cef8c25fcd14045f05d9e4fffa6ed3f62721738c056633c86ce98a3879082470R81-R93

LDiazN:
LGTM

Issue #1015, opened 2025-10-02
Add new endpoints to ooniprobe for anonymous credentials


LDiazN:
Create the new endpoints in ooniprobe for anonymous credentials

Those would be: 
- `/register` to get a credential
- `/submit` to send a measurement, this could replace `/report`
- `/update_credential` for updating a credential (not yet implemented in the library) 

Create this ONLY in the dev environment to test the client against it

PR #1016, opened 2025-10-07
Anonymous credentials draft endpoints

### LINKED ISSUE #1014: Add Python bindings as dependency for ooniprobe
Now that we have created Python bindings for the anonymous credentials protocol, we should add it as a dependency to ooniprobe 

See: https://github.com/PyO3/maturin?tab=readme-ov-file#mixed-rustpython-projects

LDiazN:
This PR adds support for the anonymous credentials protocol to the Ooni API: 

1. Adds [`ooniauth-py`](https://github.com/ooni/userauth/tree/main/ooniauth-py) as a dependency 
2. Adds  **new endpoints:** `/sign_credential`, `/manifest`, `/submit`

Note that the new measurement `/submit` endpoint won't replace the old one, and for now it's meant to be used mostly during development

Also updates the fastpath so that it's able to work with the new fields

## Migrations

This PR will require some database migrations
 
### Postgres 

Adds the following tables to postgres:
- `ooniprobe_manifest`: Describes the manifest that is reported to users when they register
- `ooniprobe_server_state`: Defines the key pair (secret_key, public_parameters) that is used for authentication

You can find the new models here: 
https://github.com/ooni/backend/blob/3b3bfb2cfe1d77a69c2e78e1f6a0067d945d429f/ooniapi/services/ooniprobe/src/ooniprobe/models.py#L58

And the migrations: 
https://github.com/ooni/backend/blob/userauth-dep/ooniapi/common/src/common/alembic/versions/7e28b5d17a7f_add_server_state_table_for_anonymous_.py

### Clickhouse

In clickhouse we need to add the fields necessary for:
- Checking  if a measurement is verified
- Running the verification offline in the fastpath machine at some point in the future when we try to implement offline verification

As an example of the changes, you can look at the `clickhouse_init.sql` script in fastpath:
https://github.com/ooni/backend/blob/3b3bfb2cfe1d77a69c2e78e1f6a0067d945d429f/fastpath/clickhouse_init.sql#L42-L46

These are the alter table statements required to run the migration in production: 

```
ALTER TABLE fastpath ON CLUSTER oonidata_cluster 
ADD COLUMN is_verified Int8,
ADD COLUMN nym Nullable(String),
ADD COLUMN zkp_request Nullable(String),
ADD COLUMN age_range Nullable(String),
ADD COLUMN msm_range Nullable(String);
```

## Feedback

This is still early work to define the API for the anonymous credentials protocol. Some things that could benefit from a bit of feedback are: 

- Naming of the endpoints
  - `/sign_credential` used to be named `/register` but it clashes with the older `/register` one, so I chose a different name
- Parameters and usage 
- Lifecycle of manifest and server state: Right now we always create a new manifest and server state if none is available to keep the API working, but we might want to be more deliberate with how we work with this
 

closes #1014 #1015 

EDIT: Adding `ON CLUSTER` to alter table query after feedback from @hellais


fastpath/fastpath/core.py
```diff
@@ -1654,7 +1654,14 @@ def process_measurement(msm_tup, buffer_writes=False) -> None:
         assert msmt_uid
         if measurement is None:
             measurement = ujson.loads(msm_jstr)
-        if sorted(measurement.keys()) == ["content", "format"]:
+
+        is_verified = g(measurement, 'is_verified', False)
```
hellais: It would be great if at some point we renamed this function `g` as something a bit more understandable. It's basically getting the key and if not it's setting to to False. Correct?

I believe modern versions of python have native support for this.

I guess we can take note of it as a future refactoring work.

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -187,10 +141,11 @@ async def receive_measurement(
...
     if content_encoding == "zstd":
         try:
+            compressed_len = len(data)
             data = zstd.decompress(data)
-            ratio = len(data) / len(data)
+            ratio = compressed_len / len(data)
             log.debug(f"Zstd compression ratio {ratio}")
-        except Exception as e:
+        except Exception:
             log.info("Failed zstd decompression")
```
hellais: Can we log `e` inside of `log.info`. Make sure you use `{e}` inside of `log.info(f"")` which should perform sanitisation.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -590,3 +565,234 @@ def random_web_test_helpers(th_list: List[str]) -> List[Dict]:
...
+    data_buff = io.BytesIO()
+    stream = io.TextIOWrapper(data_buff, "utf-8")
+    ujson.dump(data, stream)
+    data_bin = data_buff.getvalue()
+
+    # Write the whole body of the measurement in a directory based on a 1-hour
+    # time window
+    now = datetime.now(timezone.utc)
+    h = sha512(data_bin).hexdigest()[:16]
+    ts = now.strftime("%Y%m%d%H%M%S.%f")
```
hellais: Can we use ISO standard? Make sure it has in it `T` and `Z` to indicate timezone.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -590,3 +565,234 @@ def random_web_test_helpers(th_list: List[str]) -> List[Dict]:
...
+    response: Response,
+    cc_reader: CCReaderDep,
+    asn_reader: ASNReaderDep,
+    settings: SettingsDep,
+    s3_client: S3ClientDep,
+    session: PostgresSessionDep,
+    content_encoding: str = Header(default=None),
+) -> SubmitMeasurementResponse | Dict[str, Any]:
+    """
+    Submit measurement
```
hellais: Can you write a better docstring?

Also, it would be good to know where we plug in the values for "valid" `probe_age_range`s and `probe_msm_range` (can you rename these variables to that)

fastpath/fastpath/core.py
LDiazN: We can just do that I think, the fastpath now runs in python 3.12 in AWS, this can be replaced with `measurement.get('is_verified', False)` 

We can open another issue for that

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -187,10 +141,11 @@ async def receive_measurement(
     data = await request.body()
     if content_encoding == "zstd":
         try:
+            compressed_len = len(data)
             data = zstd.decompress(data)
-            ratio = len(data) / len(data)
+            ratio = compressed_len / len(data)
```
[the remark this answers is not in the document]
LDiazN: Done!

PR #1022, opened 2025-10-22
Fix bad geoip reporting

### LINKED ISSUE #1003: Bad geoip reporting
The `probe_cc_asn_nomatch_total` is returning inconsistent results. The "no match" variant is taking suspiciously high numbers when compared with the monolith version of this same metric. So we should investigate whether there's an issue with the reporting or the GeoIP db itself. 

Example (ECS version): 

Image

LDiazN:
This PR will fix the bad geoip reporting that was caused by how we parse the `X-Forwarded-For` header: 

According to MDN:
[https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#directives](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#directives)
> A proxy IP address. If a request goes through multiple proxies, the IP addresses of each successive proxy are listed. This means that the rightmost IP address is the IP address of the most recent proxy and the leftmost IP address is the address of the originating client

And AWS: 
https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html#x-forwarded-for

> The X-Forwarded-For request header may contain multiple IP addresses that are comma separated. The left-most address is the client IP where the request was first made. This is followed by any subsequent proxy identifiers, in a chain.

But we're taking the right most  header instead, the most recent proxy server (backend-fsn)
https://github.com/ooni/backend/blob/30f3860b06fd3ef22b2ceb4318043d687ecdb95e/ooniapi/services/ooniprobe/src/ooniprobe/utils.py#L127

And looking at the logs I found the following value for `X-Forwarded-For`:

`'xxx.xxx.xxx.xxx, 162.55.247.208'`

Where `162.55.247.208` is backend-fsn

So the solution should be just taking the first value instead of the last one 

closes #1003

### DIFF
--- a/ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
@@ -54,7 +54,7 @@ class Metrics:
     PROBE_CC_ASN_NO_MATCH = Counter(
         "probe_cc_asn_nomatch",
         "How many mismatches between reported and observed probe_cc and asn",
-        labelnames=["mismatch"],
+        labelnames=["mismatch", "reported", "detected"],
     )
 
     MISSED_MSMNTS = Counter(
@@ -213,7 +213,7 @@ async def receive_measurement(
 
             async with httpx.AsyncClient() as client:
                 resp = await client.post(url, content=data, timeout=59)
-            
+
             assert resp.status_code == 200, resp.content
             return ReceiveMeasurementResponse(measurement_uid=msmt_uid)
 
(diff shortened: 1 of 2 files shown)


aagbsn:
Is there an issue here where the client can inject an X-Forwarded-For header with a bogus IP? 

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#security_and_privacy_concerns

Any security-related use of X-Forwarded-For (such as for rate limiting or IP-based access control) must only use IP addresses added by a trusted proxy.

If there is a (single)  proxy as part of the ooni infrastructure, maybe it makes sense to drop the last proxy IP from the list.

Issue #1024, opened 2025-10-28
api/build_runner.sh uses deprecated debian bullseye-backports


aagbsn:
When running "integration_test" on my fork of ooni/backend, the test fails with this error:
```
 failed to solve: process "/bin/sh -c ./build_runner.sh" did not complete successfully: exit code: 100

0.478 Get:7 http://deb.debian.org/debian-security bullseye-security/main amd64 Packages [420 kB]
0.539 Get:8 http://deb.debian.org/debian bullseye-updates/main amd64 Packages [18.8 kB]

1.754 E: The repository 'http://deb.debian.org/debian bullseye-backports Release' does not have a Release file.
------
make[1]: *** [Makefile:26: .state/docker-build] Error 1
make: *** [Makefile:34: build] Error 2
make[1]: Leaving directory '/home/runner/work/backend/backend/api'
Error: Process completed with exit code 2.
```
See: https://github.com/aagbsn/backend/actions/runs/18866436670/job/53834883407

From https://backports.debian.org/Instructions/
```
When a new Debian release is made, the previous stable suite becomes oldstable. While oldstable releases may have [Long Term Support](https://wiki.debian.org/LTS), oldstable-backports are only made available for a period of one year after a new Debian stable release has been made. For example, Debian 12 'bookworm' was released on June 10th 2023, so bullseye-backports for Debian 11 'bullseye' was [closed](https://lists.debian.org/debian-backports-announce/2024/07/msg00000.html) on June 10th 2024. Packages in oldstable-backports must track the versions in stable (or stable/updates for security fixes).
```
Suggestion: switch to debian stable

PR #1026, opened 2025-10-29
Update api/build_runner.sh to debian trixie, bullseye-backports is no longer available

aagbsn:
When api/build_runner.sh is run as part of CI, it fails because debian bullseye-backports is no longer available.
I update the Dockerfile and resolve other issues with deprecated code.

### DIFF
--- a/api/tests/conftest.py
+++ b/api/tests/conftest.py
@@ -37,12 +37,13 @@ def client(app):
     with app.test_client() as client:
         yield client
 
-    while True:
-        top = flask._request_ctx_stack.top
-        if top is not None and top.preserved:
-            top.pop()
-        else:
-            break
+    # deprecated name _request_ctx_stack and marked as not a bug on issue #42
+    #while True:
+    #    top = flask._request_ctx_stack.top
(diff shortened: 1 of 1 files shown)


aagbsn:
fixes https://github.com/ooni/backend/issues/1024

api/tests/conftest.py
```diff
@@ -37,12 +37,13 @@ def client(app):
     with app.test_client() as client:
         yield client
 
-    while True:
-        top = flask._request_ctx_stack.top
-        if top is not None and top.preserved:
-            top.pop()
-        else:
-            break
+    # deprecated name _request_ctx_stack and marked as not a bug on issue #42
```
aagbsn: From https://github.com/pytest-dev/pytest-flask/issues/42, this workaround is not recommended.

api/tests/conftest.py
```diff
@@ -37,12 +37,13 @@ def client(app):
     with app.test_client() as client:
         yield client
```
aagbsn: this might need to be "return client" as well

PR #1030, opened 2025-10-29
Use safe yaml loader instead of default

LDiazN:
This PR will change the unsafe `yaml.load` in api uploader with `yaml.safe_load`


hellais (approved):
LGTM 🐳

Issue #807, opened 2024-02-23
Port public tier1 components over to the new pattern


hellais:
I started in this PR to define a nice pattern for using fastapi in conjunction with PostgreSQL for the new OONI Run v2 API: https://github.com/ooni/backend/pull/805.

I think we should adopt this for iteratively refactoring and deploying other OONI backend tier1 components.

For example the censorship findings pages should be moved over to postgresql and refactored to use fastapi.

This is also related to the API gateway work such that we are able to expose these refactored endpoints side by side the old ones minimizing the chance of breakage: https://github.com/ooni/devops/issues/7.

hellais:
Specifically for the monolith API, here is a check-list to make sure we aren't forgetting anything:

Login:

- [X] `get /api/v1/user_login`
- [X] `get /api/v1/user_refresh_token`
- [X] `get /api/v1/user_register`
- [X] `get /api/v1/login`
- [X] `get /api/_/account_metadata`
- [ ] `get /api/v1/get_account_role/{email_address}`

OONI Run:

- [X] `get /api/_/ooni_run/archive/{ooni_run_link_id}`
- [X] `get /api/_/ooni_run/create`
- [X] `get /api/_/ooni_run/fetch/{ooni_run_link_id}`
- [X] `get /api/_/ooni_run/list`

Measurements:

- [X] `get /api/v1/measurement/{measurement_uid}`
- [X] `get /api/v1/measurement_meta`
- [X] `get /api/v1/measurements`
- [X] `get /api/v1/raw_measurement`
- [X] `get /api/v1/aggregation`

Measurement upload:

- [X] `get /report`
- [X] `get /report/{report_id}`
- [X] `get /report/{report_id}/close`

Findings:

- [X] `get /api/v1/incidents/search`
- [X] `get /api/v1/incidents/show/{incident_id}`
- [X] `get /api/v1/incidents/{action}`

Check in:

- [ ] `get /api/_/check-in` (verify if this is different from the public one, see: [https://github.com/ooni/backend/blob/d943f03f829aad7e07f6bce92a8d5a6e56d68470/api/ooniapi/private.py#L1026](<https://github.com/ooni/backend/blob/d943f03f829aad7e07f6bce92a8d5a6e56d68470/api/ooniapi/private.py#L1026>))
- [X] `get /api/v1/check-in`

Probe services Other:

- [X] `get /api/v1/collectors`
- [X] `get /api/v1/geolookup`
- [X] `get /api/v1/test-helpers`
[...]

PR #1036, opened 2025-11-07
add bouncer/net-tests to ooniapi under new router bouncer

aagbsn:


ooniapi/services/ooniprobe/tests/test_bouncer.py
```diff
@@ -0,0 +1,14 @@
...
+
+def test_get_bouncer_nettests(client):
+    version = "1"
+
+    r = client.post("/bouncer/net-tests", json={"net-tests": [{"name": "foo", "version": version}]})
+    j = r.json()
+    assert "net-tests" in j
+    for v in j["net-tests"]:
+        for x in ["collector", "input-hashes", "name", "test-helpers", "test-helpers-alternate", "version"]:
+            assert x in v
```
LDiazN: Can we also port the other tests that are here: https://github.com/ooni/backend/blob/5763024ff0859fa335971080b99f6ae89766160f/api/tests/integ/test_probe_services.py#L316?

Mostly to be sure we're not breaking the API

LDiazN:
related to #807

ooniapi/services/ooniprobe/tests/test_bouncer.py
LDiazN: I would rather go with the safe route in this one, let's change it to 400.
I was looking into how this can be done and it seems like parsing the request manually is the way to go in this case, sorry for asking you to change it in another comment

We could do something like 
```
    try:
        j = await req.json()
        m = NetTestsRequest(**j)
    except pydantic.ValidationError as e:
        raise HTTPException(400, detail=e.errors())
```

(or you could manually parse the json like you had before, any choice is good to me) 

It would mess the api docs for this endpoint but in this case it doesn't really matter since this endpoint is not used by new clients and we keep it just for backwards compatibility

ooniapi/services/ooniprobe/tests/test_bouncer.py
```diff
@@ -0,0 +1,74 @@
...
+            }
+        ]
+    }
+    assert c.json() == expected
+
+
+def test_bouncer_net_tests_bad_request1(client):
+    resp = client.post("/bouncer/net-tests")
+    # XXX: fastapi returns 422 Unprocessable Entity; expected result was 400 in old test
+    assert resp.status_code != 200
```
LDiazN: As a last thing, let's also change this to `assert resp.status_code == 400` and add a comment to specify that it's for backward compatibility reasons

ooniapi/services/ooniprobe/tests/test_bouncer.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_bouncer_net_tests_bad_request2(client):
+    j = {"net-tests": []}
+    resp = client.post("/bouncer/net-tests", json=j)
+    # XXX: fastapi returns 422 Unprocessable Entity; expected result was 400 in old test
+    assert resp.status_code != 200
```
LDiazN: See previous comment: https://github.com/ooni/backend/pull/1036/files#r2533818398

aagbsn: done

Issue #1037, opened 2025-11-11
Create s3 bucket for configurations


LDiazN:
Some services in the API used to rely on configuration files to pull configuration data. We can no longer do that since we are using ECS tasks that are spawned on demand with a transient disk. We should create a private s3 bucket for configuration files that would be pulled (and cached?) on service start up

See: https://github.com/ooni/backend/pull/1033#discussion_r2502601702

Example services that might need it: 
- `/api/v1/test-list/tor-targets` : https://github.com/ooni/backend/blob/5763024ff0859fa335971080b99f6ae89766160f/api/ooniapi/probe_services.py#L673
- `api/v1/test-list/psiphon-config`: https://github.com/ooni/backend/blob/5763024ff0859fa335971080b99f6ae89766160f/api/ooniapi/probe_services.py#L659

After this is done we should go back to the following PRs: 
- https://github.com/ooni/backend/pull/1033
- https://github.com/ooni/backend/pull/1034

PR #1039, opened 2025-11-12
S3 config parameter

LDiazN:
This PR will add a setting to specify the configs s3 bucket, and an utility function to read files from the bucket

Related to https://github.com/ooni/backend/issues/1037


hellais (approved):
LGTM 🐳 🐬

PR #1052, opened 2026-01-09
make GeoLookupResult as_name field Optional with default emtpy string

### LINKED ISSUE #1051: /api/v1/geolookup returns 500 error on failure to lookup country code
curl -v  -X POST "https://ooniprobe.dev.ooni.io/api/v1/geolookup" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{  \"addresses\": [    \"1.2.3.4\"  ]}"

13:19:36.827624 [0-0] < HTTP/2 500 

This is seemingly related to a missing entry for this IP (hole in geoip dataset) does not return as_name in the response, and causes an internal server error because the result from method probe_geoip does not contain expected fields.

aagbsn:
Reserved addresses in geoip database return None for the AS Name, but default fields cc=ZZ and as=AS0. This fixes server error where AS Name was None and not validated.

Adds a test for lookups with such addresses. Fixes #1051


LDiazN:
LGTM

PR #1054, opened 2026-01-09
properly load the json

LDiazN:
Fixes error trying to read the `tor_targets.json` file from s3


ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
```diff
@@ -1,9 +1,9 @@
 import io
-from functools import lru_cache
 from pathlib import Path
 from typing import Annotated, TypeAlias, Dict, Any
 
 import boto3
+import ujson
```
aagbsn: oops!

Issue #1068, opened 2026-01-29
docker build tag is incorrect for ooniapi/services images - tag is pushed as 'v' without version


aagbsn:
see:

https://hub.docker.com/r/ooni/api-oonimeasurements/tags
https://hub.docker.com/r/ooni/api-ooniprobe/tags
vs
https://hub.docker.com/r/ooni/fastpath/tags

this might be because:
```
PKG_VERSION := $(shell hatch version)

BUILD_LABEL := $(DATE)-$(SHORT_SHA)
VERSION_LABEL = v$(PKG_VERSION)
```
whereas for backend/fastpath

```
PKG_VERSION = $(shell grep -m1 '^fastpath (' debian/changelog | sed -E 's/^fastpath \(([^)]+)\).*/\1/')


BUILD_LABEL := $(DATE)-$(SHORT_SHA)
VERSION_LABEL = v$(PKG_VERSION)
```

PR #1069, opened 2026-01-29
add pip install hatch to buildspec.yml for each ooniapi service

aagbsn:
https://github.com/ooni/backend/issues/1068


LDiazN:
LGTM

PR #1086, opened 2026-02-17
Fix/ooniprobe diff

hellais:
This addresses the diffs in this endpoint identified in the test deployment here: https://github.com/ooni/devops/pull/300

### DIFF
--- a/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
@@ -1,6 +1,22 @@
 import asyncio
 import io
 import logging
+from datetime import datetime, timezone, timedelta
+import time
+from typing import (
+    Annotated,
+    List,
+    Optional,
+    Any,
+    Dict,
+    Tuple,
+    List,
+    Any,
+    Dict,
+    Tuple,
+    Optional,
+    Annotated,
+)
 import random
 import time
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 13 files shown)


ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
aagbsn: does this field get omitted without default=None?

ooniapi/services/ooniprobe/src/ooniprobe/routers/prio_crud.py
```diff
@@ -83,19 +83,22 @@ class DebugPrioritization(BaseModel):
 )
 def debug_prioritization(
     clickhouse: ClickhouseDep,
-    probe_cc: Optional[str] = Query(description="2-letter Country-Code", default="ZZ"),
```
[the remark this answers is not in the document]
hellais: Because if it's set to `Optional` (i.e. nullable) with a default value set, effectively it's never possible for it to assume the None value, hence the correct typing should be `str`

ooniapi/services/ooniprobe/tests/test_prio.py
```diff
@@ -168,7 +169,14 @@ def test_show_countries_prioritization_csv(client):
...
+
 def test_debug_prioritization(client):
-    resp = client.get("/api/_/debug_prioritization?probe_cc=ZZ&category_codes=GOVT&probe_asn=4242")
+    resp = client.get(
+        "/api/_/debug_prioritization?probe_cc=ZZ&category_codes=GOVT&probe_asn=4242"
+    )
+    assert resp.status_code == 200
+    assert resp.headers["content-type"] == "application/json"
+
+    resp = client.get("/api/_/debug_prioritization")
```
aagbsn: see comment above; probe_cc isn't supplied in this test - so the Query parameter is effectively optional if a default value is supplied? That's why I'm surprised this test passes, because I thought Optional[] was used to indicate whether the parameter was required or not.

PR #1095, opened 2026-02-23
Optimize heavy queries

### LINKED ISSUE #1089: Investigate most heavy queries and see if constraints can be placed on them
We have a dashboard for plotting the most heavy queries against our database (see: https://grafana.ooni.org/d/b5401f39-9e48-47a9-bf87-4d8984c35fb1/slowest-clickhouse-queries?orgId=1). From here we can look into some of the ones that have the highest runtime and try to see if something can be done to either improve them or apply tighter constraints on them.

So far we have identified these:
* [ ] `aggregation` query which doesn't apply default `since` and `until` hence can lead to unbound scan without default settings: https://github.com/ooni/backend/blob/d056b84d9a2046f23f84fb7fdb1e762e2130fee2/ooniapi/services/oonimeasurements/src/oonimeasurements/routers/v1/aggregation.py#L288
* [ ] `list_measurements` which allows setting arbitrarily wide `since` and `until` time ranges. Since in this explorer view we are only really supporting loading one page of data at a time we can probably get away with limiting this to just 6 months (see: https://github.com/ooni/backend/blob/d056b84d9a2046f23f84fb7fdb1e762e2130fee2/ooniapi/services/oonimeasurements/src/oonimeasurements/routers/v1/measurements.py#L570). The performance is especially bad when a `category_code` filter is applied

LDiazN:
This PR will optimize some heavy measurements in `oonimeasurements` by changing the defaults on the API:

1. `/aggregation` now has the following defaults: `since = 6 months ago`, `until = now` 
2. `/measurements` now only allows listing measurements from 6 months ago until now

Also introduces [`freezegun`](https://pypi.org/project/freezegun/0.3.4/) to allow testing for behaviors like '6 months ago' 

Also addresses a bug in the `/aggregation` endpoint where `domain` could have the format `domain.a, domain.b` and only match `domain.a`   

closes #1089

### DIFF
--- a/ooniapi/services/oonimeasurements/tests/test_measurements.py
+++ b/ooniapi/services/oonimeasurements/tests/test_measurements.py
@@ -461,26 +461,16 @@ def test_asn_to_int():
     assert measurements.asn_to_int("1234") == 1234
 
 
-@freeze_time(FROZEN_TIME)
-def test_measurements_since_6_months_validation(client):
+def test_measurements_date_range_6_months_limit(client):
     """
-    Tests that requesting measurements with 'since' date validates the 6 months limit:
-    - Returns 400 status code when 'since' is more than 6 months ago
-    - Returns 200 status code when 'since' is within the last 6 months
+    Tests that the /measurements endpoint enforces a 6-month date range limit.
     """
-    # Date more than 6 months ago should return 400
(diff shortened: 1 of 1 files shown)


hellais (changes requested):
We should change the behaviour of the `since` and `until` restriction on listing measurements, since the main issue is requesting a large window of measurements, not necessarily that of requesting older measurements.

PR #1110, opened 2026-03-09
Add flag to deactivate metadata parsing

LDiazN:
Adds a flag to deactivate extra metadata parsing on geoip anomalies


LDiazN:
This is no longer relevant

PR #1133, opened 2026-03-20
Add psiphon config to ooniapi

aagbsn:
this API endpoint still seems to be in use; this PR stores the config file in s3 the same way as the tor_targets.json and caches it on the service instance.


LDiazN (approved):
LGTM, let's move out of backend-fsn

Issue #1070, opened 2026-01-30
Measurement anomaly detection metrics


LDiazN:
This issue will document the requirements of the measurement anomaly detection metrics we want to implement. 

# Problem

*We want to be able to detect, in real time, inconsistencies in incoming measurements*

1. Should be easy to map an inconsistency event to a measurement or set of measurements
3. We should have alerts to notice when there's a consistent increase of inconsistent measurements 
4. Should be easy to browse inconsistent measurements 
# Restrictions 

1. Identifying information about a probe user (eg. IP addresses) should ideally not be stored at all or stored for as little time as possible. Long term storage should be avoided, especially down the pipeline.
2. The resulting anomalies should be stored long term 
3. The critical path from the probe to the database should be as clear as possible, the anomaly detection system should not compromise measurement collection

# Proposed solution 

1. A Clickhouse table per type of anomaly, used to store these analytics
    - Note that every type of anomaly has different storage requirements  
3. A Grafana dashboard to browse them 
4. Some Grafana alerts based on the previous dashboard 
# Heuristics to implement

These are the heuristics we will implement. All of these require storing the `measurement_uid` of the affected measurements

## Geolocation mismatch 

- *The reported `probe_cc` is inconsistent with the IP of the probe sending the request* 
- *The reported `probe_asn` is inconsistent with the IP of the probe sending the request*  

- **How to compute it**: Compare a geolookup from the `x-real-ip` of the probe sending the measurement against the reported values 
- **Relevant data**: 
  - reported `probe_cc`, actual `cc`
  - reported `probe_asn`, actual `asn`

See: 
- https://github.com/ooni/backend/blob/c89e412796134ac80cd8a2788659aa48a1da1b05/api/ooniapi/probe_services.py#L838
- https://github.com/ooni/backend/blob/c89e412796134ac80cd8a2788659aa48a1da1b05/ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py#L161

Note that in order to compute this type of anomaly we require the probe ip which we don't store, it comes in the request and is not stored long term. So we need to: 

1. Either process the measurement in measurement collection (dangerous, but easy)
2. Store the ip just long enough to process it (safe-ish, complex)

Since we really only care about the reported and actual probe cc or asn and this comparison [is already done](https://github.com/ooni/backend/blob/c89e412796134ac80cd8a2788659aa48a1da1b05/api/ooniapi/probe_services.py#L838), we might just enqueue a task to send it to clickhouse somewhere 

[...]

PR #1138, opened 2026-03-25
Restore geoip anomaly detection

LDiazN:
This PR will restore the bad geoip faulty measurement detection that was rolled back in https://github.com/ooni/backend/pull/1109

See: https://github.com/ooni/backend/issues/1070

### DIFF
--- a/ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
@@ -167,7 +167,7 @@ def read_file(s3_client: S3ClientDep, bucket: str, file: str) -> str:
     return buff.getvalue().decode()
 
 
-async def get_tor_targets_from_s3(
+def get_tor_targets_from_s3(
     settings: SettingsDep, s3client: S3ClientDep, cache: CacheDep
 ) -> Dict[str, Any]:
     cacheKey = str(Path(settings.config_bucket, settings.tor_targets))
@@ -181,7 +181,7 @@ async def get_tor_targets_from_s3(
 
 TorTargetsDep = Annotated[Dict, Depends(get_tor_targets_from_s3)]
 
-async def get_psiphon_config_from_s3(
+def get_psiphon_config_from_s3(
     settings: SettingsDep, s3client: S3ClientDep, cache: CacheDep
 ) -> Dict[str, Any]:
     cacheKey = str(Path(settings.config_bucket, settings.psiphon_config))
@@ -193,4 +193,4 @@ async def get_psiphon_config_from_s3(
     yield resp
 
 
-PsiphonConfigDep = Annotated[Dict, Depends(get_psiphon_config_from_s3)]
+PsiphonConfigDep = Annotated[Dict | None, Depends(get_psiphon_config_from_s3)]
--- a/ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
@@ -176,7 +176,7 @@ def get_tor_targets_from_s3(
         targetstr = read_file(s3client, settings.config_bucket, settings.tor_targets)
         resp = ujson.loads(targetstr)
         cache[cacheKey] = resp
-    yield resp
+    return resp
 
 
 TorTargetsDep = Annotated[Dict, Depends(get_tor_targets_from_s3)]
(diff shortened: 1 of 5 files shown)


aagbsn:
https://github.com/ooni/backend/pull/1159 this PR uses the historical_geoip data which might help reduce the false positives as the download.db-ip.com/free/dbip-asn-lite we use now does have some holes in it. This PR consolidates the lookup_cc and lookup_asn methods into a single lookup that returns both; so we would need to refactor this PR slightly to use the single lookup method

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -942,6 +945,11 @@ async def submit_measurement(
         submit_request, manifest, settings
     )
 
+    annotations = submit_request.content.get("annotations") or {}
```
aagbsn: submit_request.content.get("annotations", {}) is a bit cleaner

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -233,6 +241,53 @@ def _set_unverified_flag(data: bytes) -> bytes:
...
+    Parse measurement body, and return the following metadata:
+
+    platform, software_name, software_version
+    """
+    try:
+        body = ujson.loads(data.decode("utf-8"))
+    except Exception as e:
+        log.error(f"Couldn't parse json body: {e}")
+        return ("", "", "")
+    content = body.get("content") or {}
```
aagbsn: same as above, can use default value with get()

ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
```diff
@@ -208,7 +208,7 @@ def read_file(s3_client: S3ClientDep, bucket: str, file: str) -> str:
     return buff.getvalue().decode()
 
 
-async def get_tor_targets_from_s3(
+def get_tor_targets_from_s3(
```
aagbsn, line 211: let's make sure this works because I'm not sure the monkeypatched tests cover this method very well

ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
```diff
@@ -217,12 +217,12 @@ async def get_tor_targets_from_s3(
...
         resp = ujson.loads(targetstr)
         cache[cacheKey] = resp
-    yield resp
+    return resp
 
 
 TorTargetsDep = Annotated[Dict, Depends(get_tor_targets_from_s3)]
 
-async def get_psiphon_config_from_s3(
+def get_psiphon_config_from_s3(
```
aagbsn, line 225: same as above with respect to monkeypatched tests not exercising this code, I think we use the async version of s3client somewhere?

ooniapi/services/ooniprobe/tests/integ/test_geolookup.py
```diff
@@ -157,15 +173,58 @@ async def test_geoip_mismatch_anoncred(client, clickhouse_db, clean_faulty_measu
...
-def check_mismatch(
-    clickhouse_db : Clickhouse,
+    # matching cc and asn
+    postj(
+        client,
+        f"/api/v1/submit_measurement/{rid}",
+        msm,
+        headers={"X-Forwarded-For": "123.123.123.123"},
+    )
+    time.sleep(0.1)  # Allow async insert to complete
```
aagbsn, line 183: hm, do we return a response before the insert has executed?

aagbsn (approved):
looks good. added a few comments.

ooniapi/services/ooniprobe/tests/integ/test_geolookup.py
LDiazN: You mean return a response to the client that made the HTTP request? Yes that's the point of this feature, to perform the analytics registration without interfering with the measurement submission path

ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
LDiazN: That method is not async, the s3 client we're using is blocking, so async it's not necessary. That's why we wrap it on a `run_on_threadpool` function instead of just `await`ing it

The only complex operation on both of these functions is this: 
https://github.com/ooni/backend/blob/e008b6fe7b5eb072b142247d3621eca7e200a36e/ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py#L217

Which in turn uses  https://github.com/ooni/backend/blob/e008b6fe7b5eb072b142247d3621eca7e200a36e/ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py#L207

And that function is blocking, not async

aagbsn: ok, thanks. I'm not sure why I added the async now

PR #1145, opened 2026-03-30
validate country codes are ISO 3166-1 alpha-2

aagbsn:
Adds pydantic_extra_types.country and validates probe_cc field in check-in.
Adds test to verify invalid probe_cc returns status 422


hellais:
Was this tested with a real client to make sure it's working as intended? If so, this looks good to merge

aagbsn:
No, this isn't tested yet with a real client. I think it makes sense to automate this type of testing by containerizing the infrastructure so that changes to backend automatically deploy and are tested against a matrix of client versions.

PR #1155, opened 2026-04-08
Remove anonymous credentials fields from the database

### LINKED ISSUE #1151: Remove zkp_request, nym from fastpath table
In light of the upcoming audit we would like to remove the following fields from the fastpath table and potentially the measurement JSON:
* zkp_request
* nym

Moroever we want to exclude these fields too since it's not possible to populate them given the current implementation:
* age_range
* msm_range

LDiazN:
This PR removes the following anonymous credentials fields from the DB: 

- `nym`
- `zkp_request`
- `age_range`
- `msm_range`

The fastpath version is also bumped

closes #1151


hellais (approved):
LGTM

PR #1156, opened 2026-04-08
Change request and response format in anonymous credentials submission

### LINKED ISSUE #1153: Add support for checking protocol version in submit_measurement
This is the backend issues for: https://github.com/ooni/probe/issues/2879

LDiazN:
This PR will update the anonymous credentials measurement submission (request and response)  formats.

Relevant changes: 

- status code 200 = measurement was processed and stored, even if verification fails
- status code 4xx or 5xx = measurement was not stored (it was rejected or there was an error processing it) 
- submission response includes error, usually related to anonymous credentials 
- The client now has to send the protocol version it has
  - If the protocol is too old a soft error is triggered: measurement is stored but not verified and the error field is set accordingly 
  - Minimum protocol version is specified as an env config variable
- The `ooniauth-py` version is updated

closes #1153


ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1015,7 +1003,90 @@ async def submit_measurement(
...
+            settings.minimum_anonc_protocol_version
+        )
+
+        if probe_version_tup < min_version_tup:
+            log.error(f"Probe version too old: {submit_request.protocol_version} < {settings.minimum_anonc_protocol_version}")
+            return False, "protocol_version_too_old", None
+
+    except Exception as e:
+        log.error(f"Unable to parse version string. probe version = {submit_request.protocol_version}, "
+        f"minimum protocol version = {settings.minimum_anonc_protocol_version}. Error: {e}"
```
hellais, line 1092: Shouldn't this be considered an error as well?

LDiazN: Yes you're right, I forgot the return in this one, good catch!

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -236,7 +238,6 @@ def probe_update_post(probe_update: ProbeUpdate) -> ProbeUpdateResponse:
 
 class CheckIn(BaseModel):
     run_type: str = "timed"
-    charging: bool = True
```
hellais, line 239: I see this was dropped from the request format, but below the default value is `False`. Do we know how the behavior was previously if `False` was being respected or `True`?

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -878,75 +903,55 @@ async def submit_measurement(
...
+        error(err_msg)
 
     try:
         asn_i = int(asn)
     except ValueError:
-        log.info("ASN value not parsable %r", asn)
-        error("Incorrect format")
+        err_msg = f"Incorrect format: ASN value not parsable {asn}"
+        log.info(err_msg)
+        error(err_msg)
```
hellais, line 933: Perhaps for consistency this could also raise the `HTTPException` directly like in the following blocks.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1021,7 +1027,94 @@ async def submit_measurement(
...
+        and submit_request.protocol_version is None
+    ):
+        return (VerificationStatus.UNVERIFIED, None, None)
+
+    # Check manifest version
+    if submit_request.manifest_version != manifest.meta.version:
+        # TODO We should validate if this is an old manifest or an unknown manifest, for now
+        # we treat them as the same error (unknown manifest)
+        log.error("Old or unknown manifest in submission request")
+        return VerificationStatus.FAILED, "manifest_not_found", None
```
hellais, line 1064: I think we should consider this as status `UNVERIFIED` since it's not an error in zkp request itself, but an issue with the client configuration.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1021,7 +1027,94 @@ async def submit_measurement(
...
+def _verify_submit(
+    submit_request: SubmitMeasurementRequest,
+    manifest: ManifestDep,
+    settings: SettingsDep,
+) -> tuple[VerificationStatus, str | None, str | None]:
+    """
+    Run the anonymous credentials verification when the relevant fields are present.
+
+    Returns (verification_status, submit_error, submit_response).
+    - "u": unverified (verification did not run)
```
hellais, line 1046: I guess we should maybe be a bit more specific with what we mean by this exactly. I'm seeing from the implementation that it's considering `unverified` only the case in which no zkp_request, nym, manifest and protocol versions are passed, but in all other cases it's either verified or failed verification.

I think it would be desirable to do it as this doc string suggests, which is to mark verified and unverified only when we actually run the zkp code and in all other case, when we aren't able to run the verification for missing or inconsistent information, we just mark it as unverified.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
LDiazN: I think it changes the size of the test list sent to the user, but the one that takes precedence is the last one, that's why I removed the first one. You can try it: 
```
from pydantic import BaseModel

class MyModel(BaseModel):
    is_charging: bool = True
    is_charging: bool = False

m = MyModel()

print(m.is_charging)

>> False
```

hellais: Ok, thanks for checking

PR #1157, opened 2026-04-09
Read ranges from policy

### LINKED ISSUE #1152: Fix ranges in submission requests
We should derive the age_range and msm_range from the manifest not have them be sent by the probe:
https://github.com/ooni/backend/blob/d4bbc4302339b4f241575f5d50ad2625be06ee93/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py#L825

LDiazN:
Reads the age range and measurement count range for ZKP verification from the manifest object

If ranges are not provided use the following defaulted ones (pending to change):  

- Age range (2461110, 2826140),
- Measurement count (0, 10000000) 

Example of a default policy: 

```
[
  {
    "match": {
      "probe_cc": "*",
      "probe_asn": "*"
    },
    "policy": {
      "age": [2461110, 2826140],
      "measurement_count": [0, 10000000]
    }
  }
]
```

Note: Due to the healthchecks, we need to update the manifest in S3 before merging this 

closes #1152


ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1116,6 +1117,59 @@ def _verify_submit(
...
+        match_asn = match.get("probe_asn", "*")
+
+        cc_ok = match_cc == "*" or match_cc == probe_cc
+        asn_ok = match_asn == "*" or match_asn == probe_asn
+
+        if cc_ok and asn_ok:
+            d = entry_policy
+            break
+
+    def_age_range, def_msm_range = get_default_ranges()
```
hellais: I don't think we should have default ranges. If the ranges are out of sync between clients and servers verification is going to be broken anyways, so it's better to fail fast than to have a default range which is broken and harder to debug.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1116,6 +1117,59 @@ def _verify_submit(
...
+
+    Matching order: first match in the list wins (highest priority first).
+
+    returns:
+    age_range, msm_range
+    """
+
+    d = None
+    for item in (policy or []):
+        if not isinstance(item, dict):
```
hellais: This should be an assert rather than skipping. If it's not a dict it means it's a malformed policy document.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1116,6 +1117,59 @@ def _verify_submit(
...
+            continue
+        match = item.get("match", {})
+        entry_policy = item.get("policy", {})
+        if not isinstance(match, dict) or not isinstance(entry_policy, dict):
+            log.error(f"Malformed entry: {item}")
+            continue
+
+        match_cc = match.get("probe_cc", "*")
+        match_asn = match.get("probe_asn", "*")
```
hellais: I think we should assume that a valid policy document MUST always have in it a catch all rule, so I would not have in here that if the key is absent it automatically becomes a catch all rule.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
LDiazN: On this line, now that we changed the format to be something more structured, what if I use a pydantic model for validation? the parsing code will be simpler (just instantiate the model) and errors will be more informative 

Also the model definition is useful for expressing the rules we're discussing: 
- Correct typing
- mandatory fields 
- no default values

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1116,6 +1117,59 @@ def _verify_submit(
 def _parse_version_tuple(version: str) -> tuple[int, ...]:
     return tuple(int(n) for n in version.split("."))
 
+def get_ranges_from_policy(
+    policy: List[Dict[str, Any]], probe_cc: str, probe_asn: str
+) -> Tuple[Tuple[int, int], Tuple[int, int]]:
```
hellais, line 1123: I think this function should assume a well structured policy document which should be validated at bootup of the service. If the policy in the manifest is not valid, the service should fail to start, rather than failover to default values.

A well formed policy document should match this criteria:

* be a list of dictionaries, where each dictionary always as the keys `match` (which always contains `probe_cc` and `probe_asn`) and `policy` (which always has `age` and `measurement_count` lists of two items)
* always contain the catch-all policy `probe_cc=*,probe_asn=*` which should be the last item in the list

Given that you assume that it's well structured, you can get rid of all the default values in the getters and conditionals on the checks.

Since this code path is security critical it's better to fail closed with an exception, than have potentially dangerous default values.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
hellais: Yeah we can use a pydantic model for that. It should be checked on service startup and we can assume it's well structured inside of this function.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1116,6 +1118,32 @@ def _verify_submit(
...
+    """
+    Gets the age and measurement count ranges from the specified policy.
+
+    Matching order: first match in the list wins (highest priority first).
+
+    returns:
+    age_range, msm_range
+    """
+
+    for item in (policy or []):
```
hellais: Since you are asserting it's a list at startup, you don't need this `or []`

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1104,8 +1106,8 @@ def _verify_submit(
             submit_request.zkp_request,
             submit_request.content["probe_cc"],
             submit_request.content["probe_asn"],
-            [2461109, 2464789],  # TODO lookup these ranges from the manifest
-            [0, 1100100100],
+            list(age_range),
+            list(count_range),
```
hellais, line 1110: Are these being cast to lists because the return value of the `age_range`, `count_range` is a Tuple? Can we make them return a list instead so we don't have to do the type conversion in here?

LDiazN: Actually this was due to a previous version of the Python API that expected a list, but I recently changed to be a tuple so we will be able to use the tuple without castings! 

It's still pending to merge but I want to merge it soon 

Also the tuple is a good way to force size in the input list from the manifest

ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
```diff
@@ -68,16 +68,52 @@ def get_cache():
...
     nym_scope: str = "ooni.org/{probe_cc}/{probe_asn}"
-    submission_policy: Dict[str, Any] = dict()
+    submission_policy: List[PolicyEntry]
     public_parameters: str
 
+    @model_validator(mode="after")
+    def _validate_catch_all(self):
+        has_catch_all = any(
+            entry.match.probe_cc == "*" and entry.match.probe_asn == "*"
+            for entry in self.submission_policy
```
hellais: This rule should always be the last rule. Having it not in the last position means that effectively everything under it will be ignored and as such should be treated as an error.

ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py
```diff
@@ -68,16 +68,63 @@ def get_cache():
...
 
     nym_scope: str = "ooni.org/{probe_cc}/{probe_asn}"
-    submission_policy: Dict[str, Any] = dict()
+    submission_policy: List[PolicyEntry]
     public_parameters: str
 
+    @model_validator(mode="after")
+    def _validate_catch_all(self):
+
+        has_catch_all = False
```
hellais:
```suggestion
        if len(self.submission_policy[-1]) < 1:
               raise ValueError("submission_policy must contain at least one element")
        # ensure the last entry in the submission_policy is the catch_all rule
        if not self.submission_policy[-1].match.probe_cc == "*" and self.submission_policy.match.probe_asn[-1] == "*":
                raise ValueError(
                 "submission_policy must include a catch-all rule with "
                "match.probe_cc='*' and match.probe_asn='*'"
                )
```

hellais (approved):
LGTM

PR #1160, opened 2026-04-11
Add more debug logging to measurement submission endpoint

hellais:
* Move probe_cc and asn comparison outside of the retry loop
* Check for 200 status code instead of asserting
* Increment send measurement failure metric inside of retry loop
* Fix bug in backup upload to s3
* Return a measurement_uid when uploading to s3 and store inside of ts/measurement_uid path


hellais:
The tests are broken, but it's currently more important to unbrick ooniprobe to avoid data loss, so I am merging it nonetheless.

In particular this fix is critical to preserve data: https://github.com/ooni/backend/pull/1160/changes/470b939adf3df5c705ddecbcb24889ec54edacac

PR #1161, opened 2026-04-11
Store files in s3 using the postcan prefix

hellais:
Storing them at these paths means we are more easily able to generate postcan files


LDiazN (approved):
LGTM

PR #1163, opened 2026-04-13
Debug fastpath

aagbsn:
add some metrics to monitor fastpath behavior more closely

### DIFF
--- a/fastpath/fastpath/core.py
+++ b/fastpath/fastpath/core.py
@@ -1600,7 +1600,6 @@ def msm_processor(queue):
 
         # already wrapped in try except inside of function body
         process_measurement(msm_tup)
-        update_fingerprints_if_needed()
 
         try:
             update_fingerprints_if_needed()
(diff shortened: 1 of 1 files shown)


fastpath/fastpath/core.py
```diff
@@ -1594,6 +1595,8 @@ def msm_processor(queue):
         process_measurement(msm_tup)
         update_fingerprints_if_needed()
 
+    metrics.decr("num_workers")
```
hellais: This should be moved below the line that says:
```
log.info("Worker with PID %d exiting", os.getpid())
```

LDiazN:
Is it possible to also log the queue size from here? 
https://github.com/ooni/backend/blob/54dec8d3fddd437316de9ffd30c021ab3eb2c393/fastpath/fastpath/localhttpfeeder.py#L40

And additionally maybe a "queue percent" metric to check how close to full the queue usually is

hellais:
> Is it possible to also log the queue size from here?
> 
> https://github.com/ooni/backend/blob/54dec8d3fddd437316de9ffd30c021ab3eb2c393/fastpath/fastpath/localhttpfeeder.py#L40
> 
> And additionally maybe a "queue percent" metric to check how close to full the queue usually is

I'm not sure we can log metrics from inside of the gunicorn worker, or what impact it would have to add logging there.

What we can do however is add a metric inside of the worker that looks up `queue.qsize()`(https://docs.python.org/3/library/queue.html#queue.Queue.qsize). I think we can also record the timing for queue.get() operations since those are blocking and we can check how much the delay in getting items from the queue are.

PR #1168, opened 2026-04-15
Increase fastpath queue size

LDiazN:
This PR will increase the fastpath queue size and reduce the worker count to match our current machine's worker count

### DIFF
--- a/fastpath/fastpath/core.py
+++ b/fastpath/fastpath/core.py
@@ -51,13 +51,22 @@ from fastpath.metrics import setup_metrics
 
 from fastpath.utils import dget_or as g_or
 
+log = logging.getLogger("fastpath")
+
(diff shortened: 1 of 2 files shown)


fastpath/fastpath/core.py
```diff
@@ -51,13 +51,22 @@
...
-LOCALITY_VALS = ("general", "global", "country", "isp", "local")
+log = logging.getLogger("fastpath")
 
-NUM_WORKERS = 12
+def _read_int_from_env(env_name : str) -> Optional[int]:
+    try:
+        val = os.environ.get(env_name)
+        return int(val) if val else None
+    except Exception as e:
+        log.error(f"Unable to read env variable '{env_name}': {e}")
```
hellais: We should throw an error on parsing it as an int, otherwise we will have a runtime error instead of failing early on startup

PR #1169, opened 2026-04-15
add obs_openvpn table to clickhouse_init.sql

aagbsn:
adds the table obs_openvpn; if testing with real data this is necessary


LDiazN (approved):
LGTM

PR #1178, opened 2026-04-17
use requests.Session to re-use connections

### LINKED ISSUE #1177: use requests.Session for connection re-use
https://github.com/ooni/backend/pull/1175 swapped out httpx library but we should be using requests.Session so that connections are re-used

aagbsn:
Use requests.Session so that connections are re-used. Also, use context manager for each response to ensure it is closed correctly even if an exception is raised or if the body is not read

fixes #1177


LDiazN (approved):
Added a commit to remove some unused imports, otherwise LGTM

PR #1179, opened 2026-04-17
Reset is_verified on submit

### LINKED ISSUE #1165: Reset anonymous credentials related fields on a measurement on submission
Our current anonymous credentials implementation works by storing the verification status along with the measurement on submission: 

https://github.com/ooni/backend/blob/bcd2979f5b093395e9f01ab00f704c048c4321b4/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py#L954

In the anonymous credentials submission path this is not an issue since the value is always overriden 

In the normal submission path however it's possible to send the field as part of the measurement body and the backend will happily store it

The task is to make sure these fields are always cleared before the measurement is processed

LDiazN:
This PR will reset the `is_verified` field on the measurement to a default value before storing it to DB in the legacy submission path 

Closes https://github.com/ooni/backend/issues/1165

### DIFF
--- a/ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
@@ -157,7 +157,7 @@ async def receive_measurement(
             error("Incorrect format")
 
     try:
-        data = await run_in_threadpool(_ensure_unverified_flag, data)
+        data = await run_in_threadpool(_set_unverified_flag, data)
     except Exception as e:
         log.info(f"Failed to parse and measurement body. Error: {e}")
         Metrics.BAD_MEASUREMENTS_CNT.labels(reason="bad_json").inc()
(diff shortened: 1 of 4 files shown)


ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -217,6 +225,14 @@ async def receive_measurement(
             return empty_measurement
 
 
+def _ensure_unverified_flag(data: bytes) -> bytes:
```
hellais: I think this should be called `_set_unverified_flag`

LDiazN: Applied!

aagbsn (approved):
this looks good to me

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -154,9 +155,17 @@ async def receive_measurement(
             Metrics.BAD_MEASUREMENTS_CNT.labels(reason="zstd_fail").inc()
             error("Incorrect format")
 
+    try:
+        data = await run_in_threadpool(_set_unverified_flag, data)
+    except Exception as e:
+        log.info(f"Failed to parse and measurement body. Error: {e}")
```
hellais:
```suggestion
        log.info("Failed to parse and modify measurement body")
        log.exception(e)
```

Issue #1088, opened 2026-02-18
Mitigate database performance issues related to heavy queries


hellais:
We are currently experiencing issues with increased database load due to a variety of different factors:
* Increase in the number of scrapers that are hitting our API endpoints at high frequency
* Some queries that have a very long runtime due to them not being particularly well tuned
* Some access patterns to explorer that can lead to very poor performance

As part of this epic we should explore ways to mitigate this issue, in light of the fact that in the worst case scenario this can have an impact on our ability to perform inserts into the clickhouse database upon measurement submission.

PR #1186, opened 2026-04-21
Return by default only a window of 7 days of data in the aggregation …

hellais:
…and list endpoints

This is related to: https://github.com/ooni/backend/issues/1088


LDiazN (approved):
LGTM 🚀

PR #1191, opened 2026-04-27
Add fastpath fallback

### LINKED ISSUE #1190: Add fallback fastpath logic in ooniprobe
Right now we only have one fastpath instance and one `fastpath_url` parameter in ooniprobe, the task is to add a fallback logic in ooniprobe to support having several fastpath instances as fallback in case a request to one of them fails

Note that oonimeasurements also needs these fastpath instances added to the `OTHER_COLLECTORS` env variable to be able to lookup raw measurements

LDiazN:
This PR will add the ability to specify a list of fastpath fallback URLs to ooniprobe. If sending a measurement fails in a fastpath instance, it will retry with the next one until all fallbacks fail or the measurement is finally sent. 

This PR will also deprecate the current `fastpath_url` setting to replace it with `fastpath_urls`. In the meanwhile, the handler endpoint will join the `fastpath_url` parameter with the `fastpath_urls` parameter to construct the list of fallbacks, but we have to remove the `fastpath_url` parameter in the future. 

Also note that oonimeasurements already supports looking up measurements from several spool dirs using the `OTHER_COLLECTORS`, we should keep them on sync in terraform  

# How to migrate

1. Merge and deploy this code
2. Deploy secondary fastpath instance 
3. Update terraform to add current fastpath instance and new one to the list of fallbacks

After this migration is applied we can safely remove the `fastpath_url` parameter 

closes #1190

### DIFF
--- a/ooniapi/common/src/common/config.py
+++ b/ooniapi/common/src/common/config.py
@@ -1,7 +1,6 @@
 from typing import Dict, List
 
 from pydantic import Field
-from pydantic.fields import Deprecated
 from pydantic_settings import BaseSettings
 
 
@@ -47,16 +46,11 @@ class Settings(BaseSettings):
     geoip_db_dir: str = "/var/lib/ooni/geoip"
     # -- < Ooniprobe only > -------------------------------------------------------------
     msmt_spool_dir: str = ""
-    fastpath_url: str = Field(
-        default="",
-        deprecated=Deprecated(
-            "This will be replaced by the `fastpath_urls` field in the near future"
-        ),
-    )  # example: http://123.123.123.123:8472
     fastpath_urls: List[str] = Field(
         description="List of fastpath instances to send measurements to",
         default_factory=list,
-    )
+    ) # example: [http://123.123.123.123:8472]
+
     failed_reports_bucket: str = (
         ""  # for uploading reports that couldn't be sent to fastpath
     )
--- a/ooniapi/common/src/common/config.py
+++ b/ooniapi/common/src/common/config.py
@@ -1,6 +1,7 @@
 from typing import Dict, List
 
 from pydantic import Field
+from pydantic.fields import Deprecated
 from pydantic_settings import BaseSettings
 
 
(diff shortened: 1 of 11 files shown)


ooniapi/common/src/common/config.py
```diff
@@ -46,7 +47,16 @@ class Settings(BaseSettings):
...
     # -- < Ooniprobe only > -------------------------------------------------------------
     msmt_spool_dir: str = ""
-    fastpath_url: str = ""  # example: http://123.123.123.123:8472
+    fastpath_url: str = Field(
+        default="",
+        deprecated=Deprecated(
+            "This will be replaced by the `fastpath_urls` field in the near future"
+        ),
+    )  # example: http://123.123.123.123:8472
+    fastpath_urls: List[str] = Field(
```
hellais, line 49: Can we always pass in a list that may contain just one element, instead of building a composite list from a single entry + the list? Having a single way to configure it is more robust to potential misconfigurations.

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -176,20 +176,29 @@ async def receive_measurement(
...
-            Metrics.SEND_FASTPATH_CNT.labels(status="ok").inc()
-            success = True
+                resp = await run_in_threadpool(client.post, url, data=data)
+                with resp:
+                    resp.raise_for_status()
+                Metrics.SEND_FASTPATH_CNT.labels(
+                    status="ok",
+                    instance=fastpath_url
+                    ).inc()
+                success = True
```
aagbsn, line 193: shouldn't this break here on success

ooniapi/common/src/common/config.py
LDiazN: Yes this is only to avoid breaking prod that depends on the `fastpath_url` parameter. We should either deploy this, update terraform to pass the correct parameters, and then remove the single entry, OR update terraform first and remove the single entry

hellais (approved):
LGTM

PR #1193, opened 2026-04-28
fix testlists /health check; do not query fastpath

aagbsn:
query citizenlab table instead of fastpath and do not query fastpath service as part of health check


LDiazN (approved):
LGTM

PR #1195, opened 2026-04-28
Remove report_id from measurement submission

### LINKED ISSUE #1139: Drop `report_id` requirement for submitting measurements
The `report_id` field requires a separate API call to open a report. And it's only usefulness is grouping measurements that correspond to multiple inputs inside an individual experiment (web_connectivity). We should phase it out to simplify the engine and clients logic.

- [ ] Create a `POST /api/v1/submit_measurement/` endpoint without `report_id` requirement in the URL
- [ ] Modify the `POST /api/v1/submit_measurement/` logic so that anonymous credentials submission is optional

LDiazN:
This PR will remove the requirement of a report id from measurement submission

1. The `/receive_measurement` endpoint will still expect the `report_id` parameter, but this will no longer be used. Instead, a pseudo-report_id will be constructed using metadata from the measurement body itself
2. The `/submit_measurement` (the new anonymous credentials version of measurement submission) will remove it entirely from the path

Note that the `report_id` will no longer be usable to group measurements that were taken in the same session, as every measurement will have a distinct `report_id`

After this PR, the `report_id` will be considered a legacy key, only kept for backwards compatibility 

A call to `POST /report`  to get the `report_id` will no longer be required before submitting measurements. 
 
closes #1139


ooniapi/services/ooniprobe/src/ooniprobe/utils.py
```diff
@@ -132,6 +133,47 @@ def generate_report_id(test_name, settings: Settings, cc: str, asn_i: int) -> st
...
+
+
+def metadata_from_measurement_content(content: dict[str, Any]) -> MeasurementMetadata:
+    """
+    Parses metadata from the `content` key in a measurement body, then
+    formats fields the same way as `generate_report_id` / `open_report` do.
+    """
+    test_name = str(content.get("test_name") or "").lower().replace("_", "")
+
+    cc = str(content.get("probe_cc") or "").upper().replace("_", "")
```
hellais: I think that since this is inside of the actual measurement content, we want to be a bit more strict about what we accept. That is to say we should enforce that the `probe_cc` is meeting all this criteria and not return a default value if it's inconsistent.

ooniapi/services/ooniprobe/src/ooniprobe/utils.py
```diff
@@ -132,6 +133,47 @@ def generate_report_id(test_name, settings: Settings, cc: str, asn_i: int) -> st
...
+    if len(cc) != 2:
+        cc = "ZZ"
+
+    raw_asn = str(content.get("probe_asn") or "AS0").upper()
+    if len(raw_asn) > 12 or len(raw_asn) < 3 or not raw_asn.startswith("AS"):
+        raw_asn = "AS0"
```
hellais: Same comment as above, the client should be setting a good value here, if they aren't we should be providing an error.

ooniapi/services/ooniprobe/src/ooniprobe/utils.py
hellais: Actually in looking at this further, I think the validation should live in the check function and in here you should just return the raw values as they as to perform validation later.

ooniapi/services/ooniprobe/src/ooniprobe/utils.py
```diff
@@ -172,10 +216,74 @@ def normalize_asn(asn: str) -> int:
...
+def check_measurement_meta(
+    test_name: str,
+    probe_cc: str,
+    probe_asn: str,
+) -> None:
+    """
+    Checks metadata consistency, raising an HTTPException and stopping
+    ingestion when an inconsistency is detected
+    """
+    asn_i = normalize_asn(probe_asn)
```
hellais: I'm not sure it's correct to be doing normalization here, since the value is coming from the measurement body. If it's not in a format compatible with the spec, we should return an error to client.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1031,19 +1000,21 @@ async def submit_measurement(
...
+            Key=s3_key,
         )
     except Exception as exc:
         log.error(f"Unable to upload measurement to s3. Error: {exc}")
         Metrics.SEND_S3_FAILURE.inc()
 
-    log.error(f"Unable to send report to fastpath. report_id: {report_id}")
+    log.error(f"Unable to send report to fastpath. measurement_uid: {msmt_uid}")
     Metrics.MISSED_MSMNTS.inc()
     return SubmitMeasurementResponse(
```
hellais, line 1026: Do we need to return to the caller the `report_id` that was generated as well? I suppose it doesn't hurt since it's a value which we generated in here and the client has no other way of knowing it.

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -242,13 +227,37 @@ async def receive_measurement(
...
+
+    assert isinstance(json, dict)
+
+    content = json.get("content")
+    assert isinstance(content, dict)
+
+    metadata = metadata_from_measurement_content(content)
+
+    json["is_verified"] = "u"
+    json["content"]["report_id"] = generate_report_id(
```
hellais: I don't think we can do this, because it will otherwise result in an inconsistency between what is inside of the measurement already (that the user got through other means) and what we are replacing it with.

In this code path, I think we should be using the `report_id` that's taken from the URL.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
LDiazN: I think the `report_id` is redundant to the `measurement_uid` in this context (it's unique per measurement). The best value we could get out of it is restoring it in the future, but in that case the probe would probably already know it before sending the measurement 

But I agree it wouldn't hurt to have add it

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
LDiazN: If we do this, we would still have to require the client to make a call to the generate report id function (or change it to generate the `report_id` on its own?). Which is fine, but we would only remove the `report_id` requirement from the anonc endpoint and not from the legacy one. Is this what we want?

hellais: The `/report/{report_id}` API call already assumes this as the report_id is encoded inside of the URL path and it cannot work without it: https://github.com/ooni/backend/pull/1195/changes#diff-fd68970e397912b47cad3b7957ef7c811249f632e065d349a0161f06fa516591R103.

I don't think we can get rid of requiring the `report_id` in a way that's backward compatible.

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
hellais: Yes I agree that it's redundant. We should check though with @aanorbel and @sdsantos if this is though still required by the mobile apps for something (I seem to recall this was still being stored in the in-app DB).

LDiazN: Norbel suggested that we keep the `report_id` in the response model, so I added that :+1:

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1026,30 +1002,34 @@ async def submit_measurement(
...
             error=submit_error,
+            report_id = rid
         )
 
     Metrics.SEND_FASTPATH_CNT.labels(status="fail").inc()
 
     # wasn't possible to send msmnt to fastpath, try to send it to s3
-    data_buff.seek(0)
+    ts_prefix = now.strftime("%Y%m%d%H")
+    tn = test_name.replace("_", "")
```
hellais: I think this is redundant, since this replacement was already done above.

ooniapi/services/ooniprobe/src/ooniprobe/routers/reports.py
```diff
@@ -157,13 +159,20 @@ async def receive_measurement(
...
     except Exception as e:
         log.info("Failed to parse and modify measurement body")
         log.exception(e)
         Metrics.BAD_MEASUREMENTS_CNT.labels(reason="bad_json").inc()
         error("Incorrect format")
 
+    # Raise an exception in case the report_id is not consistent with the body,
+    # we flag this behavior as faulty data
+    _compare_report_id_to_body_meta(cc, asn, test_name, metadata)
+    check_measurement_meta(metadata.test_name, metadata.probe_cc, metadata.probe_asn)
```
hellais, line 174: I don't think this has to be done now, but it's worth pointing out that the values of the metadata are getting validated twice. It's I think sufficient to just do a consistency check on the values in the report_id with those in the measurement metadata and then do checking of the value formats themselves in here.

This should happen naturally once the code for `/submit_measurement` and `/report/{report_id}` is refactored to share code amongst them

PR #1196, opened 2026-05-01
remove unused clickhouse-driver

aagbsn:
remove unused dependency clickhouse-driver from pyproject.toml


aagbsn:
imported by common, bah

PR #1197, opened 2026-05-05
Make only one fastpath required for healthcheck

LDiazN:
This PR will change the health check in ooniprobe to require only one fastpath to be responsive instead of every fastpath at the same time


aagbsn (commented):
looks good

PR #1198, opened 2026-05-05
add setting fastpath_timeout; requests does not timeout by default

aagbsn:
adds a configurable timeout value for posting to fastpath; by default requests does not timeout


LDiazN (approved):
LGTM

PR #1199, opened 2026-05-05
failure should be counted only on failure

aagbsn:
The metric measurement_fastpath_send_failure_count should only increment on a failure; fix indent


LDiazN:
Added a metric to count failures per fastpath instance as well

Issue #1200, opened 2026-05-08
Bugs on observations endpoint


LDiazN:
- The `observations` endpoint is returning and empty list for a valid `report_id`, example: 
https://api.ooni.org/api/v1/observations?report_id=20260413T235348Z_webconnectivity_GR_6799_n4_boA76qdHKZE0iqbi
```
curl https://api.ooni.org/api/v1/observations?report_id=20260413T235348Z_webconnectivity_GR_6799_n4_boA76qdHKZE0iqbi
{
  "metadata": {
    "count": -1,
    "current_page": 1,
    "limit": 100,
    "next_url": "https://api.ooni.io/api/v1/observations?offset=100&limit=100",
    "offset": 0,
    "pages": -1,
    "query_time": 0.33141021709889174
  },
  "results": []
}
```

- Some queries fail when you include a `probe_cc`
```
curl https://api.ooni.org/api/v1/observations\?probe_cc\='VE'
Internal Server Error
```

cc @majakomel

LDiazN:
The first one is explained by the fact that we have set the observations endpoint to have a default value for `since` and `until`, so if you just provide `report_id` like in the example, it will implicitly add a `since` and `until` range that might not contain the entries with the matching `report_id`

LDiazN:
regarding the `probe_cc`, this error is risen when querying by `probe_cc='VE'`

```
DB::Exception: Received from clickhouse1.prod.ooni.io:9000. DB::Exception: Memory limit (for query) exceeded: would use 954.64 MiB (attempt to allocate chunk of 4201806 bytes), maximum: 954.63 MiB.: (while reading column dns_engine): (while reading from part /var/lib/clickhouse/store/fbf/fbf562b8-e836-4f70-8cf0-3ba3fe2340e0/66ed98bd748b514d72ea172a58440c6e_0_3145_5/ in table ooni.obs_web (fbf562b8-e836-4f70-8cf0-3ba3fe2340e0) located on disk default of type local, from mark 301 with max_rows_to_read = 4823): While executing MergeTreeSelect(pool: ReadPool, algorithm: Thread). Stack trace:
```

It seems like this query is too memory intensive 

By running the same query again but manually in clickhouse, we get the following:


Query:
```
[...]

PR #1201, opened 2026-05-11
Fix since-until defaults over filtering query by report_id

LDiazN:
Related to #1200


aagbsn:
this looks sane to me

PR #1203, opened 2026-05-12
Update expected manifest format

### LINKED ISSUE #1202: Update manifest format
In the upcoming userauth release, we will drop the upper value of the measurement count range, see: https://github.com/ooni/userauth/pull/30#issuecomment-4268767731

Since this value will be unused, we should also remove it from the manifest spec before releasing it to the public

LDiazN:
This PR will update the expected manifest format in the backend, to remove the upper bound in measurement count range 

Also note that the validation script still works without changes because it uses the pydantic models that represent the manifest 

⚠️**important**: update the manifest in AWS immediately after merging this 

closes #1202


hellais (approved):
LGTM

PR #1212, opened 2026-06-01
Update anonymous credentials to bind submission to measurement

### LINKED ISSUE #1211: Support hash for anonymous credentials protocol
After updating https://github.com/ooni/userauth/pull/39 into `userauth`, we will have to update the measurement submission workflow to support the measurement hash binding. 

- Update the currently used version of userauth in backend 
- Update anonymous credentials submission to use the measurement hash as part of the verification process 
    - Make sure the hashing is done before altering the measurement body

LDiazN:
This PR updates the usage of anonymous credentials in measurement submission to bind the measurement verification to the measurement itself 

Now the `server_state.handle_submit` function will take the `content` key part as a string as input for the verification. The library will compute the hash internally, and use that hash to run the verification, see: https://github.com/ooni/userauth/pull/43

Note that this PR will change the request model so that `content` is a string encoding a json rather than a json itself. This is necessary to ensure hash consistency between server and client. 

Also note that we're only hashing the `content` key, rather than the entire request. This is due to the `zkp_request` field part of the metadata, which is computed using the measurement hash, it's impossible to hash the value without first having the value itself.

Another relevant change is that we're removing the extra metadata from the json blob we send to the fastpath. See: 

https://github.com/ooni/backend/pull/1212/changes#diff-ac2bf02692f757bb12106cc9be225fb135ea7aa7308e5d49155836ea1f50caa3R941

Closes #1211

### DIFF
--- a/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
@@ -1138,13 +1138,25 @@ def _verify_submit(
 def _clear_sensible_data(data : dict[str, Any]):
     """
     `data` encodes a SubmitMeasurementRequest as dict, this function
-    will create a new dict without fields with sensible information
+    will create a new dict without fields with sensible information.
+
+    The fields that should not be present in the measurement body are:
+        - zkp_request
+        - nym
+
+    These should only be used during verification, don't store them long term
     """
 
     d = {
         "content": data.get("content"),
         "format": data.get("format"),
     }
+
+    if 'protocol_version' in data:
+        d['protocol_version'] = data['protocol_version']
+
+    if 'manifest_version' in data:
+        d['manifest_version'] = data['manifest_version']
 
     return d
 
--- a/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
@@ -937,8 +937,8 @@ async def submit_measurement(
     data = submit_request.model_dump()
     data["content"] = content # change from string to dict
 
-    # Clear sensible data before sending it to fastpath
-    data = _clear_sensible_data(data)
+    # Clear sensitive data before sending it to fastpath
+    data = _clear_sensitive_data(data)
 
     # Add verification-related data.
     # use one-letter code for DB, human readable for clients
@@ -1135,16 +1135,17 @@ def _verify_submit(
         log.error(f"Unexpected anonc error: {e}")
         return (VerificationStatus.FAILED, "unknown_error", None)
 
-def _clear_sensible_data(data : dict[str, Any]):
+def _clear_sensitive_data(data : dict[str, Any]):
     """
     `data` encodes a SubmitMeasurementRequest as dict, this function
-    will create a new dict without fields with sensible information.
+    will create a new dict without fields with sensitive information.
 
     The fields that should not be present in the measurement body are:
         - zkp_request
         - nym
 
-    These should only be used during verification, don't store them long term
+    These should only be used during verification, they can leak identifying
+    data, don't store them long term.
     """
 
     d = {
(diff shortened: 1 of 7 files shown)


ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -913,10 +930,15 @@ async def submit_measurement(
...
-        submit_request, manifest, settings
+        submit_request, manifest, settings,
+        content.get('probe_cc'), content.get('probe_asn')
     )
 
     data = submit_request.model_dump()
+    data["content"] = content # change from string to dict
+
+    # Clear sensible data before sending it to fastpath
+    data = _clear_sensible_data(data)
```
hellais: Do you mean by this "sensitive" data? Looking at the function it's basically removing any extra keys that are not `content` or `format`. What are these extra fields? Can you add to the docstring what these might be and why we are removing them?

LDiazN: The sensible fields are: https://github.com/ooni/backend/blob/19b63ca50688e9acc4a76af2aef8fa93fae42d68/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py#L830-L841

(maybe we can save the manifest version and protocol version)

But yes I will add more details in the function!

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -913,10 +930,15 @@ async def submit_measurement(
 
     # Anonymous credentials verification
     verification_status, submit_error, submit_response = _verify_submit(
-        submit_request, manifest, settings
+        submit_request, manifest, settings,
+        content.get('probe_cc'), content.get('probe_asn')
     )
 
     data = submit_request.model_dump()
+    data["content"] = content # change from string to dict
```
hellais, line 938: Is there a reason for the `content` key being a JSON encoded string instead of being a dict like in the `/report/{report_id}` endpoint?

LDiazN: Yes, we need it to be a string in order to perform hashing of the `content` key. 

If it's a dict, we would have two options: 

1. Serialize it again to string and hope for the resulting string to be equal to the original json hashed by the probe 
2. Hash the entire request body as bytes, but this is not really doable due to the `zkp_request` field. It's computed using the hash of the payload, so the payload can't include the `zkp_request` (circular dependency). You can interpret it like this: `zkp_request = f(payload) = f(g(zkp_request, content, format, ...))`

hellais: Right that makes sense. Would it be better in that case to use a different key to indicate the difference in format and avoid confusion (eg. `content_str`)? Would this require a different client library?

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
hellais: Ok great, I think the correct term for that would then be `sensitive` (ie. confidential) as `sensible` means "sane or reasonable". I think this is due to a different direct translation from spanish, which is similar in italian ("sensible" means "sensitive" in italian)

LDiazN: Oh yeah, thanks for the correction, I'll change it now!

Yes in spanish we have the exact same word "sensible" for "sensitive", sorry for the confusion!

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
LDiazN: Yeah that makes sense. But yes,  this would imply changes to the client

hellais: Ok then let's not do this and keep it as-is

hellais (approved):
LGTM

PR #1214, opened 2026-06-01
Add geoip fallback to last month's DB

### LINKED ISSUE #1213: Use past month's geoip db in case of missing db
Due to how the current geoip update works right now, if the geoip is not currently published then no db will be downloaded, breaking some API services that depend on the db being present on disk: https://github.com/ooni/backend/blob/fe913ea8b5fc728aa3da7dc86594c6a600e1700c/ooniapi/services/ooniprobe/src/ooniprobe/dependencies.py#L40

This task is to use last's month db as fallback. This is not a perfect solution since we're still using an old db, but at the least we don't break measurement submission

Note that we can't use  a DB older than a month since it risks using a DB that it's too out of date

LDiazN:
This PR will add a fallback for last month's DB to the geoip download in case the service is unable to find a proper DB for this month: no new db has been published and no existent db can be used

closes #1213

### DIFF
--- a/ooniapi/services/ooniprobe/src/ooniprobe/download_geoip.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/download_geoip.py
@@ -134,6 +134,12 @@ def download_geoip(db_dir: Path, url: str, filename: str) -> None:
     Metrics.GEOIP_DOWNLOAD_TIME.observe(endtime - start_time)
 
 
+def current_geoip_db(db_dir: Path) -> Path | None:
+    if has_valid_geoip_db(db_dir):
+        return db_dir / "asn_cc.mmdb"
+    return None
+
+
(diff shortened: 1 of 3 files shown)


aagbsn (commented):
looks clean; nice work. I would note we don't appear to have a way to test the case where there isn't an update available, though. There are several cases:
db exists and there is an update
db exists and there is no update
db does not exist and there is an update
db does not exist and there is no update

I don't recall exactly what the behavior is if there is no geoip db available at all, but maybe we should make try_update return something?

LDiazN:
The previous behavior was to not download anything, so dependencies that rely on the geoip db would fail without reaching the handler code.

I guess we could refactor this to mock the download function itself to instead write an empty file to disk to test those cases

aagbsn:
thanks, this looks good

PR #1221, opened 2026-06-26
Reject ASNs with leading 0s

### LINKED ISSUE #1220: Add check to reject leading 0s in ASN
To avoid considering AS strings like `AS007` and `AS7` like the same value, we will reject ASNs with leading 0s in metadata consistency checking 

See: https://github.com/ooni/userauth/pull/45#pullrequestreview-4576260896

LDiazN:
closes #1220

### DIFF
--- a/ooniapi/services/ooniprobe/tests/test_utils.py
+++ b/ooniapi/services/ooniprobe/tests/test_utils.py
@@ -1,3 +1,5 @@
+import pytest
+from fastapi import HTTPException
 from ooniprobe.utils import check_measurement_meta, get_first_ip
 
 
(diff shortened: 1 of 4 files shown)


ooniapi/services/ooniprobe/tests/test_utils.py
```diff
@@ -7,8 +9,30 @@ def test_check_measurement_meta_http_header_field_manipulation():
...
+    assert "asn_leading_zero" in str(e.value.detail)
+
+    with pytest.raises(HTTPException) as e:
+        check_measurement_meta("web_connectivity", "IT", "AS007")
+    assert e.value.status_code == 400
+    assert "asn_leading_zero" in str(e.value.detail)
+
+    # AS0 is also invalid but for a different reason
+    with pytest.raises(HTTPException) as e:
+        check_measurement_meta("web_connectivity", "IT", "AS0")
```
hellais, line 29: Shouldn't this pass validation? Is this because we have in place the dropping of measurements with AS0?

hellais (approved):
Left a small question, but otherwise LGTM

ooniapi/services/ooniprobe/tests/test_utils.py
LDiazN: > Shouldn't this pass validation?

`AS0` didn't passed validation before, should I change this behavior to accept it?  

> Is this because we have in place the dropping of measurements with AS0?

Yes, exactly

hellais: Understood, let's leave it as-is for the moment

PR #1225, opened 2026-07-14
Fix broken geoip download tests

### LINKED ISSUE #1222: test_download_geoip fails in CI
https://github.com/ooni/backend/actions/runs/28864031192/job/85609461259#step:5:6514

```
FAILED tests/test_download_geoip.py::test_old_present_new_available - assert False is True
FAILED tests/test_download_geoip.py::test_old_present_new_unavailable - AssertionError: assert '2026-06-01' == '2026-05-01'
```
I haven't dug into this very far yet but it looks like this might be an issue that arises depending on the time that the test is run.

LDiazN:
The problem was that the decorator we use to fix the date in tests only applies to code called from within the function body. Fixtures are not affected by the decorator because they are computed before calling the decorated function

closes #1222


aagbsn:
great work, thanks for taking this on

PR #1227, opened 2026-07-20
Fix pydantic json_encoders deprecation in BaseModel

aagbsn:
Fixes deprecation warning about json_encoders


hellais (changes requested):
I think we should add a few more tests for this specific use-case. I am also seeing that some of the existing tests are failing and it's unclear if some downstream consumers might break as a result of this.

One thing we should be extra careful of is that only one serializer can be defined per field, so I'm not sure if by defining a catch all field serialiser this might impact how subclasses of this base class which use serialisers work.

Some additional tests and validation for that could help address this concern as well.

aagbsn:
Added regression tests to confirm that this PR does break backwards compatibility, and then fixed the regression.

aagbsn:
There may be some additional endpoints that need tests to ensure no other regressions exist.

aagbsn:
On this branch I rebased additional date format regression tests to verify that the services ooniauth, oonifindings, and oonirun were not affected by this change
https://github.com/ooni/backend/tree/test_datetime_regression

PR #1228, opened 2026-07-22
Profiling middleware

LDiazN:
Adds a profiling middleware that generates HTML reports of the specified endpoints, using [pyinstrument](https://pypi.org/project/pyinstrument/3.0.0b3/)

Some examples: 
image
image

This is specially useful for diagnosing performance issues, particularly the new submission path that has been showing poor performance 

# Usage 

We have now exposed two new environment variables: 
- PROFILING_ACTIVE = (default: false) Sets profiling active or unactive 
- PROFILE_REPORT_PATH = (default: empty string) The path where the HTML report will be saved to 

To specify what endpoints should be profiled, use the `whitelist` list of accepted prefixes: 
https://github.com/ooni/backend/blob/78042327342ee8759d8579a7c80f00ee321e3517/ooniapi/services/ooniprobe/src/ooniprobe/main.py#L96

### DIFF
--- a/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
+++ b/ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
@@ -1038,7 +1038,7 @@ async def submit_measurement(
 
     # wasn't possible to send msmnt to fastpath, try to send it to s3
     ts_prefix = now.strftime("%Y%m%d%H")
-    s3_key = f"postcans/{ts_prefix}/{ts_prefix}_{cc}_{tn}/{msmt_uid}.post"
+    s3_key = f"postcans/{ts_prefix}/{ts_prefix}_{cc}_{test_name}/{msmt_uid}.post"
     try:
         await run_in_threadpool(
             request.app.state.s3_client.upload_fileobj,
--- a/ooniapi/common/src/common/config.py
+++ b/ooniapi/common/src/common/config.py
@@ -106,3 +106,7 @@ class Settings(BaseSettings):
     github_token: str = ""
     origin_repo: str = ""
     push_repo: str = ""
+
+    # Profiling settings
+    profiling_active: bool = False
+    profiling_report_path: str = ""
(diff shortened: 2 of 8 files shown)


aagbsn (commented):
this looks useful. I am wondering if the ProfileMiddleware should get added via app_middleware only if profiling is enabled, rather than always adding it to the stack and doing lazy import on each request.

LDiazN:
> this looks useful. I am wondering if the ProfileMiddleware should get added via app_middleware only if profiling is enabled, rather than always adding it to the stack and doing lazy import on each request.

Yeah it would be slightly more efficient but it makes it a bit harder to test, bc then the middleware wouldn't be available on tests if the environment to activate it is not properly set in the host machine. 

A possible solution is to add a fixture that adds and removes the middleware to the `app`, what do you think?

aagbsn:
yes, if the profiling feature is intended for profiling during testing using a fixture to selectively install the middleware seems like the best solution

ooniapi/services/ooniprobe/src/ooniprobe/routers/v1/probe_services.py
```diff
@@ -1038,7 +1038,7 @@ async def submit_measurement(
 
     # wasn't possible to send msmnt to fastpath, try to send it to s3
     ts_prefix = now.strftime("%Y%m%d%H")
-    s3_key = f"postcans/{ts_prefix}/{ts_prefix}_{cc}_{tn}/{msmt_uid}.post"
+    s3_key = f"postcans/{ts_prefix}/{ts_prefix}_{cc}_{test_name}/{msmt_uid}.post"
```
hellais, line 1042: Does this impact how the s3 reprocessor will handle files?

LDiazN: No, actually the `tn` variable didn't even exist. It was a silent bug that hasn't triggered bc the fastpath has been quite stable lately. `test_name` is the new name of the same variable

aagbsn:
The changes in afd6050 look good to me; thanks for the update @LDiazN

PR #1230, opened 2026-07-27
Run verification in threadpool

LDiazN:
This PR will move verification of anonymous credentials to the threadpool

Now that https://github.com/ooni/userauth/pull/48 is merged, verifications should be able to run in parallel instead of locking the threadpool with CPU bound operations


aagbsn (approved):
great, thank you

hellais (approved):
Looks good, thanks for putting this together! 🙏

PR #1235, opened 2026-08-04
test error path of download_geoip.is_latest_available

aagbsn:
Increase code coverage towards 95% required by codecov


hellais (approved):
LGTM

hellais:
The CI is failing, I suspect because of it not being aligned with main that includes a bump to the debian runner. We should probably merge/rebase main into here and then retrigger the CI before landing this.

Issue #1238, opened 2026-08-07
bug: test-list user branch out of sync with worktree


aagbsn:
It's possible to break a user session on test-lists, where the user worktree is out of sync on filesystem with the worktree. It seems that either git.add or git.commit is failing.

hellais:
This got fixed this sprint

PR #1239, opened 2026-08-07
Fix 1238 testlists worktree

aagbsn:
This PR was made with claude sonnet-5. After reproducing a user bug ooni/backend#1238, and describing the issue, claude identified several issues. I redeployed testlists with the suggested environment variables

```
        "GITHUB_USER=ooni-bot",
        "GIT_AUTHOR_EMAIL=ooni-bot@users.noreply.github.com",
        "GIT_AUTHOR_NAME=ooni-bot",
        "GIT_COMMITTER_EMAIL=ooni-bot@users.noreply.github.com",
        "GIT_COMMITTER_NAME=ooni-bot",
```

but this did not resolve the error when a user tries to amend and resubmit a URL submission.

These changes add more verbose logging around the git.add and git.commit commands and explicitly set the contributor information.  <br>Additionally, the FileLock bug issue was detected as an error where an exception left the FileLock unclosed and this PR changes the behavior to close the FileLock explicitly.  <br>Finally, additional integration tests were made to verify the submit, modification, and resubmission workflow using a local mocked git instance which improves testing coverage of the git related methods.

### DIFF
--- a/ooniapi/services/testlists/src/testlists/manager.py
+++ b/ooniapi/services/testlists/src/testlists/manager.py
@@ -463,12 +463,11 @@ class URLListManager:
         # the user, once the PR is open the lock is acquired, when the PR is
         # closed, it's released.
         if state in ("PR_OPEN"):
-            try:
-                self._close_pr(account_id)
-            except AssertionError:
-                # This might happen due to a race between the PR being closed
-                # and it being merged upstream
-                raise CannotClosePR()
+            # _close_pr() now raises CannotClosePR() directly on failure
+            # (including the race where the PR was already merged/closed
+            # upstream between our last check and now) instead of relying
+            # on us catching an AssertionError from a bare assert.
+            self._close_pr(account_id)
             self._set_state(account_id, "IN_PROGRESS")
 
         with self._get_user_repo(account_id) as repo:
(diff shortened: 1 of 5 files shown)


aagbsn:
These tests pass locally with `make test`

ooniapi/services/testlists/src/testlists/manager.py
```diff
@@ -136,15 +161,45 @@ def get_user_lock(self, account_id: str):
...
+        # called explicitly via try/finally by every caller. Do NOT raise
+        # here: exceptions raised inside __del__ are silently discarded by
+        # Python anyway (just printed as noisy "Exception ignored in..."
+        # lines), and "the lock isn't held anymore/never was" is an
+        # entirely expected state to find here (e.g. close() already ran,
+        # or __init__ never got past get_user_lock()).
+        try:
+            self.close()
+        except Exception:
+            log.exception("[git-debug] error releasing lock in URLListManager.__del__")
```
LDiazN, line 202: will this also log the exception itself? It would be useful to get more details about it

ooniapi/services/testlists/src/testlists/manager.py
```diff
@@ -293,13 +348,34 @@ def sync_state(self, account_id) -> str:
...
+                # never revisits an account once it's CLEAN, that
+                # permanently registered the branch as "checked out" in
+                # dulwich's administrative records - the next
+                # worktree_add() for this account would raise ValueError
+                # forever, with no way to recover short of an operator
+                # manually running `git worktree prune`/`branch -D`.
+                # Leaving state as-is instead makes this self-healing: the
+                # frontend polls sync_state() every 10s while PR_OPEN, so a
+                # transient failure just gets retried on the next call.
+                log.exception(
```
LDiazN, line 372: Same note here as above

ooniapi/services/testlists/src/testlists/manager.py
aagbsn: yeah:

"log.exception() already does this by design (it's the whole reason that method exists, distinct from log.error()), and every call site in this branch is positioned correctly to benefit from it — no update needed. The only thing that's slightly redundant (harmless, not worth touching) is a few messages that also interpolate {e} into the text itself, which just means the exception's message appears twice — once in the log line, once again at the bottom of the attached traceback."

ooniapi/services/testlists/src/testlists/manager.py
aagbsn: yes it should do this by default if the call is in the right place

ooniapi/services/testlists/src/testlists/manager.py
```diff
@@ -455,16 +531,129 @@ def update(
...
     @timer(name="citizenlab_open_pr")
-    def _open_pr(self, branchname):
+    def _open_pr(self, branchname, _retried=False):
         """Opens PR. Returns API URL e.g.
         https://api.github.com/repos/citizenlab/test-lists/pulls/800
+
+        If GitHub reports a PR already exists for this branch (see
+        _pr_already_exists_error's docstring for why this can legitimately
+        happen even though this service believes no PR is open), the
+        stale PR is closed and a fresh one is opened in its place - this
```
LDiazN, line 654: Do we care about preserving the old PR? Why not reusing it? Is it because we don't know what's there?

hellais: This is a good question. It seems to me like this should be a hard failure.

I think that if two different services are pointing to the same backend and are performing operations on it, that should be something we should detect very early, because it's a deployment problem. Otherwise they will be fighting for who gets to keep an open PR and keep closing an opening each others PRs.

Are there any recoverable instances in which would happen that are not deployment misconfigurations?

aagbsn: the 'backend' in this case is github; so the development instance could be configured to point at a different upstream git repo. we don't expect users to use the test instance of test-lists alongside the production environment, but this change does ensure that if the service is out-of-sync with github, then the PR is closed and a new one opened instead of wedging.

aagbsn: the solution @hellais and I agree on is to create a new repository for the test/development instance of the test-lists service. the name of the repository should be explicitly 'test'

PR #1240, opened 2026-08-11
feat: enable webconnectivity lte in check-in response

DecFox:
This diff enables the `webconnectivity_0.5` flag in the conf.features options in check-in which allows probes to run the new webconnectivity experiment.


LDiazN (approved):
LGTM 🐳

PR #1242, opened 2026-08-13
Disable date filtering on measurement uid

### LINKED ISSUE #1233: Make observations and analysis retrievable by measurement_uid
It's currently possible to retrieve observations and analysis by measurement_uid, however we need to force the `since` and `until` fields otherwise they default to a 30 day window, which makes older observations and analysis unfetchable.

The current workaround is to just set `since=2012-01-01` so that it will search in the whole window.

And improvement could be that when you pass a measurement_uid, it should instead default to not applying a where clause on the date, so we don't have to force clients to artificially apply a very large window.

LDiazN:
* Fixes issue where analysis endpoints where applying the default date filtering even when a single measurement was requested
* Added filtering by `measurement_uid` to list observations and aggregate observations and aggregate analysis endpoints
* Added new columns to the output of list analysis: the `top_*_rule_id` family of columns

Regarding this last point: the downloaded fixtures used for testing don't have the new columns since it was taken with an older version of the DB. The current diff solves this by applying an empty string default to the table definition, but we might want to have cleaner testing data for this.

closes ooni/backend#1233


hellais (approved):
LGTM

PR #1243, opened 2026-08-17
ctrl observation endpoint

hellais:
Through this it's possible to build the measurement labeler without any new APIs and support returning these in explorer measurement pages.


ooniapi/services/oonimeasurements/Makefile
```diff
@@ -59,7 +59,7 @@ clean:
 	rm -rf .venv
 
 run:
-	hatch run uvicorn $(SERVICE_NAME).main:app
+	VALKEY_URL=memory hatch run uvicorn $(SERVICE_NAME).main:app
```
LDiazN: Can we move this to the default config? I think it's a more sensible default than the current one that assumes you have a cache service running locally

LDiazN (approved):
Just as small comment, otherwise LGTM

Issue #1244, opened 2026-08-19
testlists user branches do not update to main, automatically resolving merge conflicts is not trivial


aagbsn:
when a user submits new URLs to the test-lists service the user's git branch isn't updated to the current main. because merge conflicts do not always resolve automatically, simply merging the upstream main branch into the users branch isn't an easy fix for this problem.

currently, these PR get created and any merge conflicts are left for the reviewer to address, which has caused issues with URLs being removed accidentally.

this issue does support needing a more flexible backend for the test-lists service, e.g. using a database rather than git for the backend.

PR #1245, opened 2026-08-19
testlists: rebuild stale user branches onto master before pushing #1244

aagbsn:
Fixes merge conflicts on the PRs this service opens against citizenlab/test-lists (e.g. #2257: three commits from March, pushed and PR'd in August against a master that had moved on by five months of unrelated merged changes).

_pull_origin_repo()/_init_repo() already keep the shared repo's own copy of master fresh on every request. The actual bug is that a user's worktree/branch is only ever cut from master *once*, the first time the account starts editing (_get_user_repo()), and is never touched again no matter how far master moves before they submit.

A plain textual `git rebase` does not fix this for this workload: every edit here is a single-row append to a shared CSV (update()), and the busiest of these files (lists/global.csv, confirmed from the real repo's history) receive a steady stream of single-row-append commits from unrelated contributors. Two independent end-of-file insertions with nothing else in common are exactly what git's 3-way merge treats as a conflict - verified directly against real git, not a dulwich quirk - so replaying old commits onto a master with *any* newer, unrelated addition to the same file would still conflict, just earlier and on nearly every submission instead of rarely and on GitHub.

Instead, _rebase_user_branch_onto_master() replays the account's own tracked, structured changes (read_changes_log()/write_changes_log(), already recorded per edit) directly on top of whatever lists/<cc>.csv actually contains on master right now, then hard-resets the branch onto master and commits the result. This produces a branch that differs from master by exactly this account's pending edits and nothing else, which always applies cleanly on GitHub. It's called from _push_to_repo() right before every push.

This can raise CannotUpdateList or DuplicateURLError if replaying a change is genuinely impossible against master's current content (someone else already removed the exact row being deleted, or already added the exact URL being added) - a real conflict, not staleness. propose_changes() now re-raises any BaseOONIException from _push_to_repo() as-is instead of masking it behind the generic CannotProposeChanges, since "just retry" doesn't apply: retrying submit() alone hits the same conflict every time until the user reloads and reapplies their change. This also corrects the pre-existing uncommitted-changes-in-worktree error, which reuses CannotUpdateList, to surface as itself instead of as CannotProposeChanges.

Also fixes a latent bug in write_changes_log(): adding then deleting the same entry within one session logged a spurious delete op even though the row never reached master, which would otherwise make the replay above fail with a false "list has changed" error.

Updates test_second_users_branch_misses_first_users_merged_change (renamed to ..._picks_up_...) to assert the now-fixed behavior, per that test's own docstring, and updates
test_push_refuses_when_worktree_has_uncommitted_changes for the err_cannot_update_list vs err_cannot_propose_changes change above.

### DIFF
--- a/ooniapi/services/testlists/src/testlists/manager.py
+++ b/ooniapi/services/testlists/src/testlists/manager.py
@@ -400,12 +400,24 @@ class URLListManager:
     def write_changes_log(
         self, account_id: str, cc: str, old_entry: dict, new_entry: dict
     ):
+        """Track a single logical add/update/delete for account_id.
+
+        This log is the source of truth _rebase_user_branch_onto_master()
+        replays onto a fresh copy of master before pushing (see that
+        method's docstring), so it needs to describe the *net* effect of
+        this account's session, not just each individual update() call.
+        """
         changeset = self.read_changes_log(account_id)
         cc_changeset = changeset.setdefault(cc, [])
 
+        # True if old_entry was itself something this same session added
+        # earlier (never part of master) rather than pre-existing master
+        # content.
+        old_was_session_local_add = False
         if old_entry:
             try:
                 changeset[cc].remove(dict(old_entry, **{"action": "add"}))
+                old_was_session_local_add = True
             except ValueError:
                 # Not part of the changeset, no problem
                 pass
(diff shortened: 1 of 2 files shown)


aagbsn:
This should be reviewed for sanity since it's a pretty big changeset

PR #1247, opened 2026-08-31
feat: populate wc x flags in fastpath as part of summary upsert

DecFox:
This diff populates the `wc_x_flags` in the fastpath table as part of measurement consumer. See https://github.com/ooni/devops/pull/487

### DIFF
--- a/fastpath/fastpath/db.py
+++ b/fastpath/fastpath/db.py
@@ -80,7 +80,8 @@ def _click_create_table_fastpath() -> None:
         `engine_name` String,
         `engine_version` String,
         `test_helper_address` String,
-        `test_helper_type` String
+        `test_helper_type` String,
+        `wc_x_flags` Nullable(UInt16)
     )
     ENGINE = ReplacingMergeTree
     ORDER BY (measurement_start_time, report_id, input)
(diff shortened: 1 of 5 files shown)


fastpath/fastpath/db.py
```diff
@@ -80,7 +80,8 @@ def _click_create_table_fastpath() -> None:
         `engine_name` String,
         `engine_version` String,
         `test_helper_address` String,
-        `test_helper_type` String
+        `test_helper_type` String,
+        `wc_x_flags` Nullable(UInt16)
```
LDiazN: Why `uint16` and not a `uint32`? It affords more flexibility in the future with little memory overhead

LDiazN:
Also would be nice to have the new field added [here](https://github.com/ooni/backend/blob/f2dac67a21e392b6b5d4aef22e8a6b2c8dfe2e14/ooniapi/common/fixtures/initdb/01-scheme.sql#L5) as well

DecFox:
We decided to move `x_` flags to the `scores` dict instead of storing them in the new `wc_x_flags` since the dict is experiment specific and already stores score/analysis related data.

fastpath/fastpath/db.py
DecFox: We no longer need this since the column is no longer of use.

hellais (approved):
LGTM

LDiazN (approved):
LGTM

PR #1248, opened 2026-09-10
Update container to debian 13; debian 11 is EOL

aagbsn:
CI fails as debian 11 image is deprecated as of 2 days ago


hellais:
LGTM

PR #1249, opened 2026-09-18
Allow oonirun authors to hide their email

### LINKED ISSUE #1942: Add support for hiding email address in OONI Run links
Currently the email address of the OONI Run link creator is shown publicly. This was done for transparency reasons, but some users might have security concerns and hence we should add support for hiding it.

LDiazN:
This PR allows Oonirun link authors to hide their email from other users consuming the link.

The email will be shared by default, but authors can choose not to share their email. 

When the email is not shared, the `author` field will return `null` 

Note that admin users can always see the author email

Also the client that updates oonirun links needs to explicitly pass `share_email`, otherwise it will default to `true` and override a previous configuration 

## Migrations

This PR adds the `share_email` field to the `oonirunlink` table in **postgres**, we have to run this migration before merging to master. 

closes https://github.com/ooni/ooni.org/issues/1942


hellais (commented):
I think it would maybe be more precise to call this option `publish_email`, since the email address is still being shared with the admins. What is impacted is how it's displayed in the public endpoints.

We can possibly just implement this in the UI and not necessarily in the backend API if it's too annoying to search and replace this value.

hellais (approved):
LGTM

LDiazN:
After discussing this feature with @aanorbel, we decided to change the hidden response from `null` to the string literal `[private email]`  to avoid breaking the UI for old clients. The next client release will support this feature explicitly

PR #1250, opened 2026-09-19
feat: introduce `versions` key in check-in for experiment versions

DecFox:
This diff introduces a new `versions` key in the check-in response `conf`. This should allow the engine to pick experiment versions the engine wants to run from check-in without relying on feature flags. 

Part of OOPRIV-96


LDiazN (approved):

hellais (approved):
LGTM

PR #1251, opened 2026-09-20
Add e2e submission test with dockerized deployment

aagbsn:
Add dockerized end to end testing harness to validate end to end submission with miniooni


aagbsn:
adds test matrix to verify prebuilt and source built anonc library. the test matrix can be extended to check specific tagged versions along with master

